PR #16, opened 2024-03-25
Sonar integration

Aidasvenc:
Set up a SonarCloud integration for CI code checks.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,19 @@
+name: Sonar
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  sonarcloud:
+    name: SonarCloud
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  
+      - name: SonarCloud Scan
+        uses: SonarSource/sonarcloud-github-action@master
+        env: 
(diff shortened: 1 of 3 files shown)


Aidasvenc:
Finally it works 🥇

Aidasvenc:
Will need to set up code coverage analysis when we set up a testing library though.

.github/workflows/sonar.yml
```diff
@@ -0,0 +1,20 @@
+name: Build
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
```
AlbertoCentonze, line 7: Do we need this on `reopened` as well? When one is reopening a PR (if ever) I would assume there's not any new commit to analyze.

.github/workflows/sonar.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  sonarcloud:
+    name: SonarCloud
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
AlbertoCentonze, line 13: Please update this to the latest version of checkout (v4)

AlbertoCentonze (changes requested):
Nice job! We'll probably need to customize the settings a bit more as we use it but for now it's a great start.

.github/workflows/sonar.yml
Aidasvenc: Copied from the suggested code in the docs. You are right.

AlbertoCentonze (approved):
LGTM

PR #24, opened 2024-04-01
feat: contact list

GaelCondeLosada:
### What I did
I developed a screen that allows users to view all their contacts in a list format. This screen displays each contact's profile picture, name, type, and a brief description. Additionally, there's a feature for users to search specific contacts. The screen includes a button to toggle between "Plain View" (the current list view) and "Graph View." However, the Graph View is still under development, so the button is non-functional at present.

### How I did it

- I created an interface 'Contact' struct.
- To facilitate testing, I created mock contacts data, pending the integration of a functional database.
- I utilized standard React Native components to implement the user interface.

### Result

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/cc201189-5d2a-48d8-8b99-7cb22768d631

### Notes

- The contacts are currently dummy data.

- Contacts are not yet sorted alphabetically. There's the possibility of sorting them dynamically, but I'm concerned about potential efficiency issues. Pre-sorting at load time or ensuring they are initially loaded in a sorted manner might be more effective, and I'm open to discussing this.

- The top and bottom navigation bar are not implemented because we'll make a component for the whole app. Also, the bottom navigation bar rises when using the keyboard, but I haven't addressed it yet as it will be replaced.

- I thought of implementing both the list view and the graph view in the same file and just use an "if" statement to choose which one to display. We can also use navigation but this will need a few changes in the way the two buttons "Plain View" and "Graph View" are coded.


assets/colors/colors.ts
```diff
@@ -1,2 +1,8 @@
-export const white = 'white'
-export const peach = '#FD9890'
\ No newline at end of file
+export const white = 'white';
+export const peach = '#FD9890';
+export const red = 'red';
+export const blue = 'blue';
+export const black = 'black';
+export const lightgray = '#E2DFDE';
```
Aidasvenc, line 6: Please use camell-case for variable namings.
```suggestion
export const lightGray = '#E2DFDE';
```

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,124 @@
...
+        setFilteredContacts(filtered);
+    }else{
+        setFilteredContacts(dummyData);
+    }
+  }
+
+  const TopScreen = () => { // will be a module
+    return (
+      <View style={styles.horizontalContainer}>
+        <Image
```
Aidasvenc, line 52: You probably missed this when we talked about this in-person, but importing a `.png` or a `.svg` for an icon is not efficient. You should install and use `@expo/vector-icons` package which provides you with many icons out of the box. Also, all of the icons are stylable. You can find all icons listed [here](https://icons.expo.fyi/Index). E.g. you would probably want to use https://icons.expo.fyi/Index/Ionicons/location-outline from Ionicons here.

1. Import it in your file with `import { Ionicons } from '@expo/vector-icons'`
2. Use it and style the icon as you wish by rendering `<Ionicons name="location-outline" size={24} color="black" />`

Aidasvenc (changes requested):
Wow, a really great job for now. I really like your style of coding, very neat. There are just some minor details that you need to fix and you will be good to go. 🚀

screens/Contacts/ContactListScreen.tsx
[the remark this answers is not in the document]
GaelCondeLosada: I agree but I think I'll leave it like that for the moment and we'll make it an argument when we have the data fetched from the data base if it's ok with you.

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,125 @@
...
+        )
+        setFilteredContacts(filtered);
+    }else{
+        setFilteredContacts(dummyData);
+    }
+  }
+
+  const RenderOneContact = ({ item } : { item : Contact }) => (
+    <View style={styles.cardContainer}>
+      {item.profilePictureUrl? (
```
GaelCondeLosada, line 52: Here I'm still using an image and an icon for the default "no profile picture". I believe this is the way to go.

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,125 @@
+import { useState } from 'react';
+import { View, Text, FlatList, Image, TextInput, TouchableOpacity} from 'react-native';
+import { styles } from './styles';
+import { Ionicons } from '@expo/vector-icons'
```
alexandre-mrt, line 4: Be sure to stipulate that in your PR next time. To be aware of the new dependencies.

screens/Contacts/ContactListScreen.tsx
alexandre-mrt: Agree, Ionicons is the way to go

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const RenderOneContact = ({ item } : { item : Contact }) => (
+    <View style={styles.cardContainer}>
+      {item.profilePictureUrl? (
+        <Image
+          style={styles.profilePicture}
+          source={{uri: item.profilePictureUrl}}
+        />
+      ) : (
+        <View style={styles.profilePicture}>
+          <Ionicons name="person" size={50} color="black" />
```
alexandre-mrt, line 59: add the color to the dedicated file when you can :)

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </View>
+      </View>
+    </View>
+  )
+
+  return (
+    <View style={styles.container}>
+      
+      <View style={styles.horizonatlBarContainer}>
+        <Text>Here will be the top navigation bar</Text>
```
alexandre-mrt, line 80: I will implement that in a futur branch

screens/Contacts/ContactListScreen.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </View>
+
+      <View>
+        <TextInput 
+          style = {styles.searchBar}
+          placeholder="Search..." 
+          value={searchText}
+          onChangeText={handleSearch}
+        />
+      </View>
```
alexandre-mrt, line 106: for the simple <View> without styling you can use <></> it more concise (but it's small details)

screens/Contacts/styles.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+    profilePicture: {
+        alignItems: 'center',
+        borderColor: peach,
+        borderRadius: 40,
+        borderWidth: 2,
+        height: 80,
+        justifyContent: 'center',
+        // this is needed to make the border of the profile picture and the botder of the contact overlap
+        left: -2,
```
alexandre-mrt, line 71: take care to negative values, it can cause trouble

alexandre-mrt (commented):
Very good work, only remarks are micro-optimisations or guidelines

PR #30, opened 2024-04-08
feat: force-graph improvements

gaspardthrl:
For now the graph is draggable in every direction, zoomable, nodes can be dragged and moved individually or simply clicked which will result in a console.warn.

Implemented dummy Images and names for nodes to track them.

To display the graph:

```
const graph = new Graph(['0', '1', '2', '3', '4', '5', '6'], ['0', '2', '3', '3', '3', '4', '5', '6', '6', '6'], ['1', '1', '1', '2', '0', '1', '1', '1', '2', '0'], [0.5, 0.2, 0.2, 0.5, 0.5, 0.5, 0.2, 0.5, 0.5, 0.5]);
const constrainedNodeId = '1';

<ForceDirectedGraph graph={graph} constrainedNodeId={constrainedNodeId}/>
```

https://github.com/uniconnect-epfl/uniconnect/assets/81919237/20ced7ce-6393-4537-8efb-6fe185791d14


graph/ForceDirectedGraph/ExampleForceDirectedGraph.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Graph from '../Graph';
+import ForceDirectedGraph from './ForceDirectedGraph';
+
+const ExampleForceDirectedGraph = () => {
+    const graph = new Graph(['0', '1', '2', '3', '4', '5', '6'], ['0', '2', '3', '3', '3', '4', '5', '6', '6', '6'], ['1', '1', '1', '2', '0', '1', '1', '1', '2', '0'], [0.5, 0.2, 0.2, 0.5, 0.5, 0.5, 0.2, 0.5, 0.5, 0.5]);
```
AlbertoCentonze, line 6: Would rather have this in a local README.md file that you can copy and paste from whenever you need to test this. As long as we don't have proper unit tests it's better not to have test code.

graph/ForceDirectedGraph/ForceDirectedGraph.tsx
```diff
@@ -0,0 +1,310 @@
...
+          dy: 0,
+        }
+      )}
+      y2={coordY(
+        nodes.find((node) => node.id === link.target) ?? {
+          x: CENTER_WIDTH,
+          y: CENTER_HEIGHT,
+          id: link.target,
+          dx: 0,
+          dy: 0,
```
AlbertoCentonze, line 205: Might be nice to create a helper function to create objects of this type, maybe one accepting only `x`, y` and `id` and defaulting to zero whenever `dx` and `dy` are not specified.

AlbertoCentonze (changes requested):
Good job overall. Nice is very well written.

graph/ForceDirectedGraph/ForceDirectedGraph.tsx
gaspardthrl: I was planning to create a specific Node Class instead of an interface to account for these kinds of methods during the next milestone (when the database would be linked).

So I think it's good for now.

Tell me what you think.

AlbertoCentonze (approved):

PR #51, opened 2024-04-09
Gesture handler requires correction on android

GaelCondeLosada:
### What I did
Fixed an error when launching the app on an

WhatsApp Image 2024-04-10 at 00 17 01_1ffc76dd

### How I did it

I had to add the dependency `"react-native-gesture-handler": "~2.14.0"`.
However, when I first added a previous version of the dependency (the "1.10.3") and this got me another error:

WhatsApp Image 2024-04-10 at 00 17 01_22f80436

Upgrading the dependency to version 2.14.0 got rid of the error.

I then also wrapped the app into a GestureHandlerRootView as recommended [here](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/installation/). However this didn't fix the issue, the actual fix was to add the dependency and import it, I tried with and without the GestureHandlerRootView. I kept it because it seemed good practice for our application configuration.

### Result

The app works now normally on my android device

### Note 

- Please don't forget to test the app on IOS during review.
- Because it helped me with my problem I think we should address the warnings when we'll have the time:

warnings

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -2,6 +2,8 @@ import { NavigationContainer } from '@react-navigation/native';
 import React, { useEffect } from 'react';
 import * as SplashScreen from 'expo-splash-screen';
 import RegistrationStackNavigator from './navigation/Registration/RegistrationStackNavigator';
+import { GestureHandlerRootView } from 'react-native-gesture-handler';
+import styles from './styles';
 import {
   useFonts,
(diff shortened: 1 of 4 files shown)


Aidasvenc (approved):
Thank you for spotting and fixing the problem! good job!

PR #52, opened 2024-04-09
test: add unit tests for current codebase

GaelCondeLosada:
### What I did

Added the unit testing library "jest" and made unit tests for all currently present app code.

### How I did it

Added necessary dependencies:
- `jest`
- `@types/jest`
- `jest-expo`
- `@testing-library/react-native` for testing utilities

Added a script so you can just run `npm run test`.
I modified the .gitignore to exclude all coverage feedback files.
I made unit tests for all the current functional screens we have
**Note that the tests don't cover everything yet on some files** (I didn't have time to polish everything) **and everyone will also have to produce sufficient unit tests every time a new functionality or fix is pushed from the point this branch is merged.**
Made contact data a property for the contact list but this is a detail.

### Result

Screenshot_12-4-2024_11155_

### Notes

- There are still warnings when running jest (which isn't easily fixable, I tried doing what is said):

Capture

- As you can see in results, two files are not well covered. App content is not meant to be a part of the app, it was useful for development only so I didn't test it. For the Login screen, I can't mock Firebase which prevents me from testing the majority of the screen. We should use dependency injection for Firebase here.

### DIFF
--- a/__test__/screens/Contacts/ContactListScreen.test.tsx
+++ b/__test__/screens/Contacts/ContactListScreen.test.tsx
@@ -1,37 +1,16 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react-native'
-import { ContactListScreen, Contact } from '../../../screens/Contacts/ContactListScreen'
+import { ContactListScreen } from '../../../screens/Contacts/ContactListScreen'
 
 describe('ContactListScreen', () => {
-    const dummyData: Contact[] = [
-        {uid: '1', firstName: 'Jocović', lastName: 'Dellpouelo', profilePictureUrl: '', description: 'This guy is very weird', interests: ['surfing', 'machine learning'], qualifications: []},
-        {uid: '2', firstName: 'Stephano', lastName: 'Carasto', profilePictureUrl: '', description: 'No description', interests: [], qualifications: ['surfing', 'machine learning']},
-        {uid: '3', firstName: 'Henrique', lastName: 'Nique', profilePictureUrl: '', description: 'Good question', interests: ['surfing'], qualifications: ['machine learning']},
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -44,7 +49,22 @@
     "eslint-plugin-react-native": "^4.1.0",
     "eslint-plugin-yml": "^1.12.2",
     "husky": "^9.0.11",
+    "jest-expo": "^50.0.4",
```
Aidasvenc, line 52: Was this automatically added as a dev dependency?

GaelCondeLosada: I belive so yes. I followed these instructions: https://docs.expo.dev/develop/unit-testing/. Should it be only in normal dependencies ?

Aidasvenc: It is possible that it is needed as a dev. dependency too. You can check by removing it from the dev. dependencies list and then doing an `npm install`. If the tests still execute without issues, it means that the package is not needed as a dev. dependency.

GaelCondeLosada: Ok, works fine wihtout it. Removed it.

__test__/screens/Contacts/ContactListScreen.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react-native'
+import { ContactListScreen } from '../../../screens/Contacts/ContactListScreen';
```
GaelCondeLosada, line 3: Imported from the same file twice

Aidasvenc (approved):
LGTM!

Aidasvenc:
It seems that you changed the list file to not contain dummy data by default. In order to have a non-empty flow in this screen, you should leave the dummy data as is for now.

PR #61, opened 2024-04-11
fix: improved design and logic for M1

Fyuryy:
# Fixing design and adding some logic 

What was done?
- Replaced Interest grid by FlatList component for performance reasons
- Fixed textbox bug that didn't allow the user to click on the boxes
- Added feature allowing the user to quit the keyboard by clicking on the screen
- Added feature allowing the user to go to the next input box by clicking 'Enter'
- Corrected selection colors in the interest selection list
- Fonts load before any screen is displayed

What is there to do during the next milestone?
- Add date picking feature to allow the user to input their birth date in an easier and more intuitive way.
- Interactions with Fire base so that users can create accounts properly 
- Code the rest of the screens for the 
<p float="left">

# Unselected and selected interests

</p>


GaelCondeLosada (changes requested):
I didn't have the time to check all the code in details but there are a few things that would require changes in my opinion.

- First of all I suggest you can delete the file `AppContent.tsx` as it is not a part of the app.

- Secondly on Android there are a few problems:
After clicking on Add a description now, the button still says "Add a description now" and the bottom navigation bar is slightly cut out. However I think the navigation bar being slightly cut out is not a top priority problem to fix.


Another problem that's a bit more concerning is when trying to type a description, the keyboard goes over the description section.

screens/Onboarding/OnboardingScreen.tsx
```diff
@@ -13,66 +21,76 @@ const OnboardingScreen: React.FC = () => {
...
-        placeholder="Username or email"
-        placeholderTextColor={'black'}
-        keyboardType="email-address"
-        autoCapitalize="none"
-      />
+        {/* Username/Email Field */}
+        <TextInput
+          style={[styles.inputField, globalStyles.text]}
+          placeholder="Username or email"
+          placeholderTextColor={'black'}
```
GaelCondeLosada, line 37: I think you should use the colors.ts file in assets for the colors.

screens/Registration/AuthenticationScreen/AuthenticationScreen.tsx
```diff
@@ -49,83 +56,85 @@ const AuthenticationScreen: React.FC = () => {
...
+        <InputField
+          label="Confirm Password*"
+          placeholder="****************"
+          onChangeText={setConfirmPassword}
+          value={confirmPassword}
+        ></InputField>
+
+        <View style={styles.container}>
+          <View style={styles.phrase}>
+            <Entypo name="cross" color="red" />
```
GaelCondeLosada, line 84: Same thing for the colors. There are other ones too.

App.tsx
```diff
@@ -18,28 +18,43 @@ export default function App() {
...
     async function prepare() {
-      await SplashScreen.preventAutoHideAsync();
+      try {
+        await SplashScreen.preventAutoHideAsync();
+        await new Promise((resolve) => setTimeout(resolve, 2000));
+      } catch (e) {
+        console.warn(e);
+      } finally {
+        setAppIsReady(true);
+      }
```
AlbertoCentonze, line 32: Can this fail? Do we need a try/catch block here?

components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <Pressable style = {styles.absView} onPress={() => setDateModal(false)}>
+      <View style={styles.dateContainer}>
+        <DateTimePicker
+          // minimumDate={}
+          // maximumDate={}
+          testID="dateTimePicker"
+          value={date}
+          mode="date"
+          display={Platform.OS === 'ios' ? 'spinner' : 'default'}
```
AlbertoCentonze, line 37: Was this properly tested on android?

components/InputField/InputField.tsx
```diff
@@ -1,33 +1,36 @@
...
   return (
     <View style={styles.section}>
-      <Text style={[styles.topText, globalStyles.text]}>{label}</Text>
+      <Text style={[styles.label, globalStyles.text]}>{label}</Text>
       <TextInput
         placeholder={placeholder}
         style={[styles.input, globalStyles.text]}
         value={value}
         onChangeText={onChangeText}
+        onSubmitEditing={onSubmitEditing}
```
AlbertoCentonze, line 29: Do we need `forwaredRef` for this to work?

screens/Registration/InterestsScreen/InterestsScreen.tsx
```diff
@@ -41,7 +44,7 @@ const InterestButton: React.FC<InterestButtonProps> = ({
 
 const interests = [
   'Machine Learning',
-  'Artificial Intelligence',
+  'Artificial Intelligence is very cool',
```
AlbertoCentonze, line 47: I think this was a test but you forgot to undo it ;)

screens/Registration/InterestsScreen/InterestsScreen.tsx
```diff
@@ -53,6 +56,11 @@ const interests = [
   'Ethereum',
   'Solana',
   'Computer graphics',
+  'Bananas',
+  'Apples',
+  'Big tech',
+  'Finance',
+  'Gaspard tes teubé t con',
```
AlbertoCentonze, line 63: Please remove this

AlbertoCentonze (changes requested):
Overall a very nice PR, let's try to make more smaller PRs next time as we discussed during the retrospective! Also I would recommend already removing all the semi-columns everywhere since this will come up as a conflict during merge.

App.tsx
Fyuryy: I think it's better if we let it for safety reasons

components/DatePicker/DatePicker.tsx
Fyuryy: yes!

components/InputField/InputField.tsx
Fyuryy: forwardRef is necessary when you want to pass a ref down to a child component that is not directly accessible due to being encapsulated within another component. It allows the child component to expose its own ref to the parent component. Without forwardRef, the ref you pass to the child would reference the instance of the child component itself, not the underlying DOM element or native control you're trying to access.

screens/Registration/InterestsScreen/InterestsScreen.tsx
Fyuryy: Done

AlbertoCentonze (approved):
LGTM!

GaelCondeLosada (approved):
Looks pretty good to me now, the layout is still not perfect with the keyboard on android but the functionality work fine. Date picker is functional on android. It is good enough for me we ca make another PR to fix small details later.

PR #68, opened 2024-04-12
Implement registration and login logic in sync with a database

GustaveCharles:
### What I did

I merged the Login logic and the Onboarding Screen

### How I did it

I separated the logic and the states. The logic is in the Login, and the states in Onboarding Screen. 
I still have to finish setting up a load screen, although this would need to be done once and used at each time.

### Small video

https://github.com/uniconnect-epfl/uniconnect/assets/91184289/2091faa5-181a-4669-88ee-642e56d9f488


GaelCondeLosada (approved):
LGTM

PR #75, opened 2024-04-15
Profile screens components

GaelCondeLosada:
# What I did
Added two components to render the basic of a profile and its description. These components will be used to make the profile screens later. The goal is to modularize them because the different profile screens will be very similar.

# How I did it
Defined two new components:
- ExpandableDescription
- GeneralProfile

Implemented them with react native components and made their unit tests in two new test files. I modified MyProfileScreen.tsx to display them for the moment. I didn't add any new dependency nor modified any other file.

# How to verify it
Launch the app and log in. You can see the two components displayed in the "my profile" screen (by clicking on the top left of the screen).

# Demo 

Capture

Capture

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/components/GeneralProfile/styles.ts
+++ b/components/GeneralProfile/styles.ts
@@ -1,0 +1,30 @@
+import { StyleSheet } from "react-native"
+import { peach } from "../../assets/colors/colors"
+
+export const styles = StyleSheet.create({
+    container: {
+        justifyContent: 'center',
+        left: 10,
+        margin: 10,
+    },
+    horizontalContainer: {
+        flexDirection: "row"
+    },
+    location: {
+        fontSize: 11,
+    },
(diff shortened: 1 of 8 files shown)


components/GeneralProfile/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+    profilePicture: {
+        alignItems: 'center',
+        backgroundColor: peach,
+        borderRadius: 40, // to make it circular
+        height: 80,
+        justifyContent: 'center',
+        width: 80,
+    },
+
+})
```
Fyuryy, line 30: Stick to one type of quotes, you are using both ' and " in this file.

Fyuryy (approved):
LGTM

PR #78, opened 2024-04-15
feat: implement google authentication

GustaveCharles:
# What I did
 I implemented the login functionality with google. I have also implemented tests for login with firebase and with google

# How I did it

- I added onto the actual google login screen, the logic behind the call to google API
- To make the login work I installed a new `@react-native-google-signin/google-signin` library and connected it with firebase.

# How to test it

When running the code, you can print the "user" const in OnboardingScreen.tsx and see that it returns a token as well as the other user values from google.

Testing with Expo Go does not work for this feature, thus you should start your own local simulator (with Expo Orbit for example) and download an APK. [This APK](https://expo.dev/accounts/uniconnect-epfl/projects/uniconnect/builds/88c246fc-aa61-433a-b803-749246ab8c91
) is already a valid one to test the functionality 

# What has been unit tested?

I have tested the Onboarding Screen functionalities. I have mocked the `loginEmailPassword` and the `GoogleSignin.signIn()` functions, which means that the backend hasn't been tested. This is done on purpose so that backend and frontend testing is separated. 

Thus, an upcoming PR should be to test the backend functionalities thoroughly. 

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/91184289/c37c75a5-5e17-4687-a852-5275f6c9e1d1

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


package.json
```diff
@@ -16,14 +16,16 @@
     "@expo/metro-runtime": "~3.1.3",
     "@expo/vector-icons": "^14.0.0",
     "@react-native-community/datetimepicker": "^7.6.4",
+    "@react-native-google-signin/google-signin": "^11.0.1",
     "@react-navigation/bottom-tabs": "6.5.20",
     "@react-navigation/native": "6.1.17",
     "@react-navigation/stack": "6.3.29",
-    "eas": "^0.1.0",
     "@testing-library/react-native": "^12.4.0",
+    "eas": "^0.1.0",
```
Aidasvenc, line 24: why did you install eas? I think it is a mistake

app.config.js
Aidasvenc, line 1: why did you rename the file?

Aidasvenc (commented):
Thank you for the functionality! However, I feel like some important context details are missing in the PR description. Please describe what library you used and why you needed to hange config files. Also, you should mention that the google login does not work in expo go, it works only when built. It would be nice to know how to test it externally. Thanks!

package.json
```diff
@@ -4,8 +4,8 @@
   "main": "node_modules/expo/AppEntry.js",
   "scripts": {
     "start": "expo start",
-    "android": "expo start --android",
```
GaelCondeLosada, line 7: This changes the behaviour of the app in expo go, is it on purpose ? Here is what chatGPT has told me on the difference between start and run :

    Development Environment vs. Native Environment:
        expo start: Runs the app within the Expo Go app, which contains pre-built native code and APIs provided by Expo.
        expo run: Builds and runs the app with your specific configurations and native code, directly on the emulator, simulator, or device.

    Quick Iteration vs. Detailed Testing:
        expo start is optimal for developers looking to quickly make changes and test outcomes in a dynamic development environment.
        expo run is crucial for testing more complex integrations and performance issues that can only be identified in a standalone app environment.

    Managed vs. Bare Workflow Features:
        While both commands can be used in a managed workflow with Expo, expo run steps closer to what is traditionally considered a bare workflow by directly interacting with the native codebase.

app.config.js
```diff
@@ -15,9 +18,11 @@
       "**/*"
     ],
     "ios": {
-      "supportsTablet": true
+      "supportsTablet": true,
     },
     "android": {
+      
+      "googleServicesFile": process.env.GOOGLE_SERVICES_JSON,
```
GaelCondeLosada, line 25: Is this only needed on android ?

GaelCondeLosada (commented):
I have an error when running your branch on my phone :

Capture

I don't have time yet but I'll look more into it if I get the time. (This is not a problem with `npm install`, I ran it twice to be sure).

And not all the unit test pass on my computer, but I believe it's just because you didn't pull the last version of main.

app.config.js
GustaveCharles: Yes it was to export the google-services.json, which is a secret file on the local device

app.config.js
GustaveCharles: Yes, as we are only doing the build on Android

package.json
GustaveCharles: You're right thanks

Aidasvenc (approved):
Good job with the tests! Must be a nice feeling to have learned to mock all needed modules 👏

GaelCondeLosada (approved):
Good job ! Just don't forget to address sonar cloud issues before merging.

AlbertoCentonze:
Hey Gustave, I see you haven't checked the `works on android` box. Do you need any help with that?

PR #82, opened 2024-04-16
feat: added persistent storage

GustaveCharles:
# What I did
I implemented persistent storage following Adais' remark

# How I did it
I followed [this](https://stackoverflow.com/questions/76914913/cannot-import-getreactnativepersistence-in-firebase10-1-0) Stack Overflow thread and I added this dependency '@react-native-async-storage/async-storage'

# How to verify it

# Pre-merge checklist
- [ ] work correctly
- [ ] are fully tested

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,8 @@
 {
-  "extends": "expo/tsconfig.base",
   "compilerOptions": {
-    "strict": true
-  }
+    "paths": {
+      "@firebase/auth": ["./node_modules/@firebase/auth/dist/index.rn.d.ts"]
+    }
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
-  "extends": "expo/tsconfig.base",
   "compilerOptions": {
-    "strict": true
```
Aidasvenc, line 4: It seems that you have by accident deleted `"strict": true`

Aidasvenc (commented):
One minor fix is needed and you will be good to go!

PR #86, opened 2024-04-16
feat: added DescriptionScreen

Fyuryy:
Now the description screen is separated from the information screen

# What I did
Created the Description Screen, its respective styles file and tested it 
Made changes to Authentication Screen and Information Screen and their respective styles and test files. 
Tested the DatePicker component
 
# How I did it
Created a new folder called <b>DescriptionScreen</b> with a DescriptionScreen.tsx file and a style.ts file.
Used jest to test the respective screens and components.

# How to verify it
Follow the registration process and at the <b>InformationScreen</b> click on <i>Add description now</i>
Click inside the box to open the keyboard and click outside of it to close it.
Click on the red icon to go to the <b>InterestsScreen</b> as intended.
There's no interaction to the Database yet in this screen.

# Demo video
https://github.com/uniconnect-epfl/uniconnect/assets/91114548/1da9dc17-0159-4945-8911-511032761469

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


screens/Registration/DescriptionScreen/DescriptionScreen.tsx
```diff
@@ -0,0 +1,55 @@
...
+        </Text>
+
+        <TextInput
+          style={[styles.description, globalStyles.text]}
+          multiline={true}
+          textAlignVertical="top"
+          onChangeText={setDescription}
+          value={description}
+          placeholder="Enter your description here..."
+        ></TextInput>
```
GaelCondeLosada, line 43: You can just close the text input with />, no need to repeat it

GaelCondeLosada (commented):
Overall I think the code is good, not much to say on that. And everything works fine on my phone. I think there are a few things we can discuss on the design .

On the information screen I think it would be a bit more intuitive if the navigation would stay at the bottom, this is also the way this screen was made in the figma.

Capture

On the description screen there are a few things that feel weird to me:
- The "validate" button icon, I prefer the way this button was in the figma.
- The border width is larger than in the rest of the app, I don't know if this is intentional
- When validating the description we go forward in the registration with no feedback. I think I would feel better as a user if it would make me go back to the information screen and I would have to validate all the informations.

Capture

screens/Registration/DescriptionScreen/DescriptionScreen.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+        <TouchableOpacity
+          onPress={() => navigation.navigate("Interests" as never)}
```
Aidasvenc, line 46: Instead of navigating forward, you should call `navigation.goBack()` since conceptually after entering a description you should go back to the main screen to verify the information you entered and only then navigate forward

Fyuryy: Will do!

screens/Registration/DescriptionScreen/DescriptionScreen.tsx
Fyuryy: I will definitely change the validation button, as I dislike it too.. Thanks for the other points.

GustaveCharles (approved):
LGTM. Good work Pedro. One thing to note, is that you changed a lot of files, which was maybe a bit out of scope of the PR.

PR #87, opened 2024-04-17
feat: my profile screen

GaelCondeLosada:
# What I did
Implemented a part of the "my profile screen". The lower part with events and interests is not implemented yet. I intend to implement it in another PR (or someone else can do it). The screen is usable nonetheless.

# How I did it
- Implemented the screen in MyProfileScreen.
- Slightly modified GeneralProfile and ExpandableDescription component to make them easier to use.
- Created MyQrCodeScreen and UpdateMyProfileScreen so we can navigate to them from the MyProfileScreen
- Added tests for MyProfileScreen, GeneralProfile and ExpandableDescription

# How to verify it

You can go to "my profile" by clicking on the top left of the screen. You can also run the tests.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/ca86691b-d9f0-4894-a1af-9457a1f5a90d

# Pre-merge checklist
The changes I have introduced:
- [x] work correctly
- [x] do not break other functionalities
- [x] work correctly on Android
- [x] are fully tested

PR #89, opened 2024-04-17
Component for selectable choices

GaelCondeLosada:
# What I did

# How I did it

# How to verify it

# Demo video

# Pre-merge checklist
- [ ] work correctly
- [ ] are fully tested

### DIFF
--- a/components/SelectableChoices/SelectableChoices.tsx
+++ b/components/SelectableChoices/SelectableChoices.tsx
@@ -1,0 +1,52 @@
+import React, { useState } from "react"
+import { View, Text, TouchableOpacity } from "react-native"
+import { styles } from "./styles"
+
+interface SelectableChoicesProps {
+  choices: string[],
+  startingChoice: string,
+  selectBarWidth: number,
+  onChoiceChange: (newChoice: string) => void
+}
+
+const SelectableChoices: React.FC<SelectableChoicesProps> = ({
+    choices,
+    startingChoice = choices[0],
+    selectBarWidth = 80,
+    onChoiceChange = () => {}
+}) => {
+
+    const [currentChoice, setCurrentChoice] = useState(startingChoice)
+  
+    const RenderOneChoice = ({ choice }: { choice: string }) => (
+        <TouchableOpacity
+            style={styles.choiceContainer}
+            onPress={() => {
+                setCurrentChoice(choice)
+                onChoiceChange(choice)
+            }}>
+                
+            <Text 
+                style={choice === currentChoice ? styles.selectedText : styles.unselectedText}>
+                {choice}
+            </Text>
+            <View 
+                style={[
+                    choice === currentChoice ? styles.selectedBar : styles.unselectedBar, 
+                    {width: selectBarWidth}
+                ]}
+            />
+
+        </TouchableOpacity>
+    )
+
+    return (
+        <View style={styles.choicesContainer}>
+            <View/>
+            {choices.map(choice => RenderOneChoice({choice}))}
+            <View/>
+        </View>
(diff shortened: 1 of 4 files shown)


components/SelectableChoices/SelectableChoices.tsx
Aidasvenc, line 1: A reusable component is a component that you can then re-use multiple times. It seems like you defined a list of such components in the `components` folder. I would rather refactor the file into `SelectableChoice` and make it return one `SelectableChoice`. Also, the name is a bit weird, I suggest something something that sounds more like a component and defines better what it is: `SelectableSection` or even just `Section`

components/SelectableChoices/styles.ts
```diff
@@ -0,0 +1,31 @@
...
+        justifyContent: "space-between",
+    },
+    selectedBar: {
+        backgroundColor: peach,
+        borderRadius: 2,
+        height: 4,
+    },
+    selectedText: {
+        color: black,
+        fontWeight: "bold",
```
Aidasvenc, line 21: Same as in #87. Please start using our custom font.

screens/Contacts/ContactListScreen.tsx
```diff
@@ -242,13 +226,8 @@ export const ContactListScreen = () => {
...
       </View>
 
-      <View style={styles.container}>
-        <FlatList
-          data={filteredContacts}
-          renderItem={RenderOneContact}
-          keyExtractor={(contact) => contact.uid}
-        />
-      </View>
+      {selectedTab === "Plain View" ? RenderContactList() : RenderContactGraph()}
```
Aidasvenc, line 229: If a graph computation will be a resourceful computation, we may have to cache the graph view so that it is not re-rendered on each switch between two sections. It would mean that both these components will have to become [top tabs](https://reactnavigation.org/docs/material-top-tab-navigator). Let's keep it like that for now though:)

Aidasvenc (changes requested):
Minor refactorings are needed and you will be good to go!

components/SelectableChoices/SelectableChoices.tsx
GaelCondeLosada: Hey, I agree with you on the name, I'm kind of struggling to find a good one. I changed it to "SectionTabs", tell me if this one feels better to you. As for the fact that this is more a list of components than a component itself, I get what you're saying but I'll try to provide you a few argument for why I did it this way.

- We'll be using them a few times in the app but every time in the exact same way of having one to three tabs. This makes it easier for others to use it without having to bother with correct layout and updating.
- I don't think we'll ever be using the singular tab component outside of this specific "list" usage.
- Using it, I feel more intuitive to write :
`<SectionTabs 
        tabs={["Plain View", "Graph View"]}
        startingTab="Plain View"
        onTabChange={tab => {setSelectedTab(tab)}}
      />`
than :
`<View style={styles.tabsContainer}>
        <SectionTabs
          name="Plain View"
          selected={selectedTab === "Plain View"}
          onPress={setSelectedTab("Plain View")}
        />
        <SectionTabs
          name="Graph View"
          selected={selectedTab === "Graph View"}
          onPress={setSelectedTab("Graph View")}
        />
      </View>`

Tell me if this made you change your mind. If you still think it would be a better idea to just make a component for a singular tab, I'll change it.

Aidasvenc: I agree with you now! Thanks for your explanation

Aidasvenc (approved):
LGTM now!

AlbertoCentonze:
Please correct/acknowledge sonar cloud issues before merging!

PR #99, opened 2024-04-19
feat: external profile screen

GaelCondeLosada:
# What I did
- Implemented the screen to see other people's profile. 
- In the profile screen implemented the part to see the contact's network.
- I also added the screen in the navigation and slightly modified the contact list to be able to navigate to the profile screen when clicking on a contact in the contact list. 
- For the time being, all the profile and network data is all dummy data.

# How I did it
- Implemented the profile screen in `ExternalProfileScreen.tsx`
- Made tests for the external profile screen.
- For this profile screen I put the majority of the styles in a `profileStyles.ts` file in the `Profile` folder because they are the exact same as for the MyProfile screen and I intend to also use these style when refining this screen (MyProfileScreen) to minimize code duplication (Tell me what you think about it).
- I defined three "components" in the `Profile` folder: `ProfileEvents` `ProfileInterests` and `ProfileNetwork`, because they are a part of the screen that repeat for MyProfileScreen and ExternalProfileScreen so I didn't want to duplicate them. I didn't want to put them in the `component` folder because they are not component but more a "half screen". I don't know if it's the best place to put them or the best way to do it, tell me what you think about it.
- I made tests for these three "components".
- I slightly modified the contact list so each contact is a button on which you can click to see their profile. (Note that I also had to export the dummy data to the ProfileNetwork because I didn't want to create new data and Sonar cloud didn't let me copy paste it and make "code duplication". It's a bit ugly but I think it's ok, it'll go away soon...)
- I had to also modify the contact list tests to reflect the changes.

# How to verify it
Launch the app with expo go, go the the contact list and click on a contact. You can also run the unit tests.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/72e911c2-c95d-4476-8571-836d61120b12

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


screens/Contacts/ContactListScreen.tsx
```diff
@@ -167,7 +169,9 @@ export const ContactListScreen = () => {
   }
 
   const RenderOneContact = ({ item }: { item: Contact }) => (
-    <View style={styles.cardContainer}>
+    <TouchableOpacity 
+      style={styles.cardContainer}
+      onPress={() => useNav.navigate({name: "ExternalProfile", params: {uid: item.uid}} as never)}>
```
gaspardthrl, line 174: Why are you using `as never`?

I think that `never` is supposed to be used in very precise situations, this one not being one of them.

[See here for more information](https://stackoverflow.com/questions/72001800/can-someone-explain-to-me-what-does-as-never-do-in-this-piece-of-code)

screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
```diff
@@ -1,10 +1,104 @@
...
+  qualifications: string[],
+  location: string
+}
+
+const dummyProfile: Contact = {
+  uid: "4",
+  firstName: "Hervé",
+  lastName: "DelaStrite",
+  profilePictureUrl: "",
+  description: "Description, description, descrc hdsjklaf hel fhj fdj bjhd vbfdhj vbdjhl vdjlh vdhj vfdjh hvdu vdh vdfj vdfs hfdj fdj vdfjl d grei hrj heiua fjispd jis gjdrkl gjreks reksl grei gril nrsj njiption, loooooooooong descriiiiiiiiiiption, this is a veeeeeeeeeeeeeeeeeeeeeery loooooooooooooooooooooooooooooooooooooooooooong descriiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiption alala",
```
gaspardthrl, line 29: I would advise using _Lorem Ipsum_ for this kind of dummy data as it will make visual checks clearer and more relevant.

[See here for Lorem Ipsum generator](https://loremipsum.io/)

gaspardthrl (changes requested):
I submitted a couple changes. Beside that, great work!

screens/Contacts/ContactListScreen.tsx
GaelCondeLosada: You're right, I didn't look into it but just replicated some existing code. I modified it. Note that we are using as never quite a lot when navigating and I guess this is a mistake.

Capture

gaspardthrl: Yes, I think that here people use it for type inference but worst case scenario you should use `as any`.

gaspardthrl (approved):
Great work

PR #120, opened 2024-04-24
feat: screen scan qr

GaelCondeLosada:
# What I did
I added a screen from which you can scan use the camera to scan QR codes. This screen will start by asking camera permissions if the app doesn't have them. We can discuss the UI because this screen wasn't in the Figma. Also note that there is a bit of latency when changing home screen. I tried my best to reduce it by changing screen before the camera mounts or unmounts but I think there is room for improvement.

# How I did it
- Added dependency `expo-camera` in `package.json`.
- Implemented the screen in `QrScanScreen.tsx` and added it to the navigation.
- Added adequate tests for the screen.

# How to verify it
You can **test the authorization button** when arriving on the connection screen (from home navigation). If the app does not ask for authorization, you can remove camera permissions in settings for expo go (without closing the app expo go) and relaunch uniconnect, this should make it ask you for permission. After testing permissions, you can **test the camera and try to scan a QR code** and see if the right alert pops. You can also run the tests.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/794f6cf7-f576-42e8-9b6a-1a5bebc6d6de

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/__test__/components/TabBar/TabBar.test.tsx
+++ b/__test__/components/TabBar/TabBar.test.tsx
@@ -37,17 +37,15 @@ describe("TabBar", () => {
     const connections = getByText("Connections")
     const explore = getByText("Explore")
 
-    fireEvent.press(connections)
-
-    expect(getByText("Plain View")).toBeTruthy()
-
     fireEvent.press(home)
 
-    expect(getByText("Home")).toBeTruthy()
+    expect(getByText("Show on the Map")).toBeTruthy()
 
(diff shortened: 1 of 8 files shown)


AlbertoCentonze (approved):
Neat PR! Totally fine for this sprint, but you should consider "stop" qr code recognition once the first one has been recognized to avoid the alert spamming you have shown in the demo!

PR #121, opened 2024-04-25
feat: linked graph view with navigation

gaspardthrl:
# What I did
I linked the graph view with the rest of the app.
I implemented some basic navigation and search functionalities.
I also extracted part of the code from ForceDirectedGraph.tsx and implemented it in another file to have better modularity over components.

# How I did it
I added the component inside the ContactGraph.tsx file and added some arguments to ForceDirectedGraph to implement navigation logic.

# How to verify it
1. Run the application
2. Go in the Explore section
3. Go in the Graph View mode
4. Switch to the plain view mode
5. Switch again to the graph view mode, the layout should not have changed given that you did not modify the drag by dragging or scrolling.
6. Try to interact with the graph
7. Typing in the search bar numbers from 0 to 10 included should "highlight" corresponding nodes
8. When clicking return on the keyboard, if at least one node is selected, you should be redirected to an external profile screen dummy screen
9. When keyboard is up and you click outside of it, the keyboard should disappear
10. When clicking on a node an animation should start and a modal should appear
11. Clicking on the node's profile picture displayed on the modal should redirect to an external profile screen dummy screen

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/81919237/cd1f8b63-6446-45d6-aeba-3321ebf7e431

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


GaelCondeLosada:
Don't forget to address sonar cloud issues.

gaspardthrl:
> Don't forget to address sonar cloud issues.

Yes, don't know why I didn't see those earlier...

Thanks!

screens/Contacts/ContactGraph/ContactGraph.tsx
```diff
@@ -1,12 +1,95 @@
...
+
+function handleSearch(text: string, graph: Graph): void {
+  if (text === "") {
+    for (const node of graph.getNodes()) {
+      node.selected = false
+    }
+  }
+  else {
+    for (const node of graph.getNodes()) {
+      if (node.id.includes(text) || text.includes(node.id)) {
```
GaelCondeLosada, line 77: Here you had a TODO saying to use regex for text search. I think this should be implemented sometimes. Maybe you can add an issue for that.

GaelCondeLosada (commented):
Overall good work ! I requested a few changes in the code. I also have a few suggestions on the UI.
First I think it could be good to push the modal upwards so we can still use the navigation bar when having a modal on screen: 



We'll need the usage of regex, but will probably implement it later: 

WhatsApp Image 2024-04-25 at 17 08 07_5d5f47b0

The graph goes under the present UI. I think this was done on purpose but it personally makes me feel a bit weird (it's maybe just me). Here is the screen I got when opening graph view for the first time: 

WhatsApp Image 2024-04-25 at 17 08 07_a2895ae4

Aside from that everything is good and everything works on my android device !

screens/Contacts/ContactGraph/ContactGraph.tsx
gaspardthrl: Yes I had one but eslintrc did not let me keep it

components/Graph/ForceDirectedGraph/ForceDirectedGraph.tsx
```diff
@@ -96,23 +95,30 @@ const ForceDirectedGraph: React.FC<{
   // Use effect to update the graph
   useEffect(() => {
```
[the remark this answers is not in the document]
gaspardthrl: Almost certain that it does not as it is only called at the begining or when we update the actual graph's data

components/Graph/Graph.tsx
```diff
@@ -160,6 +179,14 @@ export default class Graph {
...
       return link.source !== source || link.target !== target 
     }) 
+    this.setInitialized(false)
+  }
+
+
+  getNodeById(id: string): Node | undefined {
+    return this.nodes.find((node: Node): boolean => {
+      return node.id === id 
+    })
```
AlbertoCentonze, line 189: Please when dealing with nullable types always please prefer `null`. `undefined` is for unexpected nullable values while `null` is for intentional ones.

components/Graph/graphAlgorithms/FruchtermanReingold.tsx
```diff
@@ -20,14 +20,14 @@ export const fruchtermanReingold = (
   iterations: number
 ) => {
   // Maximum distance between nodes
-  const k = Math.sqrt((width * height) / nodes.length)
+  const k = 100
```
AlbertoCentonze, line 23: Why was this set to 100 instead of dynamic computation?

screens/Contacts/ContactGraph/ContactGraph.tsx
```diff
@@ -1,12 +1,95 @@
...
+import { useState } from "react"
+
+import NodeModal from "../../../components/Graph/NodeModal/NodeModal"
+
+const graph = new Graph(
+  ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10"],
+  ["0", "2", "3", "4", "5", "6", "7", "8", "9", "10"],
+  ["1", "1", "1", "1", "1", "1", "2", "2", "3", "3"],
+  [0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5]
+)
```
AlbertoCentonze, line 15: Please don't use example code in the final codebase. You can maybe create a separate file called `mocks.ts` for the component so that testing parts remain isolated and can be easily deleted when no longer necessary.

AlbertoCentonze (changes requested):
Very nice PR.

As more research bars stack-up in the project we will probably need to add some debouncer to avoid having our phones lagging when tapping on the keyboard. Especially in your case this should probably be prioritary for the next sprint!

components/Graph/Graph.tsx
gaspardthrl: Can't since this.nodes.find returns Node | undefined

AlbertoCentonze (approved):

PR #125, opened 2024-04-26
fix: resolved fallback problem on App and ScrollView on Event Screen,…

alexandre-mrt:
Resolved bugs that blocked the demo

# What I did
I deleted the fallback option on the App screen that blocked the navigation for the HomeScreen
I added a back button on the Event map to go back to the previous screen. I added the styling too and tests for it
I resolved a but on the ScrollView in HomeScreen that prevented to scroll smoothly.

# How I did it
I deleted the fallback option as only the redirection to the HomeTabNavigation was necessary. 
I added the button "back-arrow"
I deleted the Touchable opacity that was hiding the keyboard, as it conflicted the ScrollView.

# Demo video

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/components/EventCard/EventCard.tsx
+++ b/components/EventCard/EventCard.tsx
@@ -28,9 +28,9 @@ const EventCard: React.FC<EventCardProps> = ({
         <Text style={styles.title}>{title}</Text>
         <Text>{date}</Text>
         <View style={styles.locationContainer}>
-          <Text style={styles.location}>{location}</Text>
-          <Text>{latitude}</Text>
-          <Text>{longitude}</Text>
+          <Text style={styles.location}>{location +  "  "}</Text>
+          <Text>{Number(latitude).toFixed(2) + "  "}</Text>
+          <Text>{Number(longitude).toFixed(2)}</Text>
         </View>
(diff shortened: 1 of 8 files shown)


components/EventCard/EventCard.tsx
```diff
@@ -28,9 +28,9 @@ const EventCard: React.FC<EventCardProps> = ({
         <Text style={styles.title}>{title}</Text>
         <Text>{date}</Text>
         <View style={styles.locationContainer}>
-          <Text style={styles.location}>{location}</Text>
-          <Text>{latitude}</Text>
-          <Text>{longitude}</Text>
+          <Text style={styles.location}>{location +  "  "}</Text>
+          <Text>{Number(latitude).toFixed(2) + "  "}</Text>
+          <Text>{Number(longitude).toFixed(2)}</Text>
```
gaspardthrl, line 33: Good for now but I think we will need to use at least 3 decimal points since 2 decimal points gives us an error radius of 2.78 km I think.

gaspardthrl (approved):
Added a single comment, also next time try to import colors from style. Otherwise LGTM

PR #126, opened 2024-04-26
feat(contact): new contact structure and on-device storage

gaspardthrl:
# What I did
- I implemented the contact type and refactored all code requiring this structure.
- I implemented a simple, asynchronous, unencrypted, persistent, key-value storage system for storing the graph object to avoid recreating one each time. Note that the graph object is deleted at the launch of the app and re-created the first time we navigate to the __Explore__ section.

# How I did it

- I created a Contact type in Contact.tsx and refactored files involved in the __Explore__ section which were all using their own dummy data.
- I used the [AsyncStorage API ](https://react-native-async-storage.github.io/async-storage/)
- I extracted the methods from the Graph class as AsyncStorage can only store __strings__ and functions cannot be **serialized** making them impossible to use when retrieving the Graph object from the storage system.

# How to verify it

To check the AsyncStorage:
1. Navigate to the __Explore__ section and click on the __Graph View__
2. Check that the graph layout does not change when navigating to another part of the app and coming back to the __Graph View__ (given that you did not modify the drag by dragging it or any node)
3. Reload the application, navigate to the __Explore__ section and click on the __Graph View__
4. Check that the graph layout is different from last time (Note that it can still be the same as the node placement algorithm can converge to the same positions as last time - in this case retry this step one or two times)

To check the Contact implementation:
1. Check that contacts, either in the __Plain View__ (i.e., list) or in the __Graph View__ have their own profile picture and that they are displayed correctly
2. In the __Graph View__ check that the further a node is from the **main node** (i.e., the user itself), the smallest it is.

# Demo video
https://github.com/uniconnect-epfl/uniconnect/assets/81919237/4c4a5933-bbe7-499c-974a-29788b95748b

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


App.tsx
```diff
@@ -16,6 +16,30 @@ import * as WebBrowser from "expo-web-browser"
...
+      // Reset the flag to indicate that the graph file no longer exists
+      await AsyncStorage.setItem(GRAPH_EXISTENCE_FLAG_KEY, "false")
+    }
+  } catch (error) {
+    console.error("Error destroying graph file:", error)
+  }
+}
+
+// Call the function to destroy the graph file when the app launches
+destroyGraphFileIfExists()
```
GustaveCharles, line 41: shouldn't this function be in another file? Ideally we do not want function implementations in App.tsx

GustaveCharles (commented):
LGTM, very good implementation and harmonization of the overall mocking of the database. The database is coming, and it will be way easier to implement it with your job already done. Thanks a lot!

App.tsx
```diff
@@ -16,6 +16,30 @@ import * as WebBrowser from "expo-web-browser"
...
 // import HomeScreen from "./screens/Home/HomeScreen"
 
+import AsyncStorage from "@react-native-async-storage/async-storage"
+
+const GRAPH_STORAGE_KEY = "graph" // Key used by AsyncStorage to store and retrieve the graph file
+const GRAPH_EXISTENCE_FLAG_KEY = "graph_exists" // Key used by AsyncStorage to store a flag indicating whether the graph file exists
+
+// Function to destroy the graph file if it exists
+const destroyGraphFileIfExists = async () => {
+  try {
```
GustaveCharles, line 26: shouldn't this function be in another file? Ideally we do not want function implementations in App.tsx

GustaveCharles (changes requested):

App.tsx
```diff
@@ -16,6 +16,30 @@ import * as WebBrowser from "expo-web-browser"
...
+    if (graphExists === "true") {
+      // If the graph file exists, delete it
+      await AsyncStorage.removeItem(GRAPH_STORAGE_KEY)
+      // Reset the flag to indicate that the graph file no longer exists
+      await AsyncStorage.setItem(GRAPH_EXISTENCE_FLAG_KEY, "false")
+    }
+  } catch (error) {
+    console.error("Error destroying graph file:", error)
+  }
+}
```
GustaveCharles, line 38: shouldn't this function be in another file? Ideally we do not want function implementations in App.tsx

GustaveCharles (changes requested):

App.tsx
gaspardthrl: You are right, I defined it in the ExploreScreen.tsx now

components/Graph/NodeModal/NodeModal.tsx
```diff
@@ -1,78 +1,67 @@
...
-]
-
+/**
+ * NodeModal component
+ * @param node - The node to display
+ * @param visible - Whether the modal is visible
+ * @param onPressOut - Function to call when the modal is pressed out
+ * @param onContactPress - Function to call when the contact is pressed
+ * @returns The NodeModal component
+ */
```
Aidasvenc, line 20: I don't think it is really necessary to add documentation for components - especially when the prop variable names are already intuitive enough

gaspardthrl: Okay good points!

screens/Contacts/Contact.tsx
Aidasvenc, line 1: To be fair, I have some mixed feelings regarding the creation of a class. Even though I fully respect your way of `Java OOP` thinking, I am not sure whether it would be particularly useful for our project structure. The problem is exactly that which arises in classical OOP - not all `Contacts` are the same, for some contacts you should be able to see and query their friends, for others you don't. Moreover:

1. Defining a constructor and getters or setters is just plain boilerplate code which has no practical meaning in javascript + typescript codebases. Anyone can still call `Contact.uid` or `Contact.friends` and there is no way to restrict it like you would do with private or public methods.
2. Ideally, all database querying/updating functions will be defined and exported in their dedicated files under `firebase/` folder.

Our main objective is to have a seamless integration between database (backend) and front-end. Therefore, in my view, we need to focus to have unified model definitions between frontend and backend. This is why in my PR I have defined a new folder with global types, where each type used there is mimicking a model returned by the database. The database functions themselves are such that they already take care of handling errors so that these errors do not get passed to the user-side as blocking operations, but rather as functions that could fail silently (user's actions do not get restricted by a thrown error, user is only informed that something has failed and asked to try again).

It may still be a point of discussion, but in my overall view, we should not aim for classes, we should focus on types only.

gaspardthrl: Ok, I will reimplement it through a class

screens/Contacts/ContactGraph/ContactGraph.tsx
```diff
@@ -1,88 +1,131 @@
...
+    getNodeById(graph, userId)
+  )
 
+  // State to store whether the modal is visible
   const [modalVisible, setModalVisible] = useState(false)
 
+  // Function to call when the modal is pressed
   const onModalPress = (uid: string) => {
-    const node = graph.getNodeById(uid)
+    const node = getNodeById(graph, uid)
```
Aidasvenc, line 46: Here we have the same problem - commenting this type of code does not bring value, the variable/function names are already intuitive enough. One good example which clearly did need comments was the `ForceDirectedGraph` file, where not so-intuitive functions were defined. I am not sure whether the comments are needed here

screens/Contacts/ExploreScreen.tsx
```diff
@@ -1,45 +1,118 @@
...
+          )
+
+          await AsyncStorage.setItem(GRAPH_EXISTENCE_FLAG_KEY, "true")
+          // Set the graph state
+          setGraph(newGraph)
+        }
+      } catch (error) {
+        console.error("Error loading graph data:", error)
+      }
+    }
```
Aidasvenc, line 49: There you have a perfect example of where comments are needed - a function that contains many if statements and its' internal structure is not trivial to understand without comments

Aidasvenc (changes requested):
A very nice job! Your graph implementation is progressing very fast!

I have some doubts about class definitions and value of comments, but it's just a minor discussion point to address.

GustaveCharles (approved):

Fyuryy (approved):
Overall your code is very readable and understandable. However, the size of the PR might be to big. Maybe you could divide it into a couple of smaller ones, making one PR per task (or multiple similar tasks). For example one PR for graph handling, one PR for the contacts handling and one for restructuring previous files. 
That said, it is frankly impressive to see the amount of work you get done, and how well you get it done. Good job Gaspard!

PR #127, opened 2024-04-27
Add user data

Aidasvenc:
# What I did
This is a big one. It covers finalizing a complete User registration flow and storing/fetching all User data to/from Firestore Database.

- Redefined database fetching/registration functions to be non-blocking. All error codes now have a mapping to user-friendly error messages.
- Instead of an alert, a `Toast` message is now displayed. This helps the flow to be less blocking (no need to dismiss an alert).
- All information filled out during the registration phase is now saved using `Context` and passed during the creation of a user model in the database.
- Real user data is now displayed in the profile screen.
- Each database model will now have its' global type that will be defined in `types` folder.

One detail that we will need to address during our meeting is how to refine the Google Sign-in registration flow. Currently, if a user does a first time sign-in, its' user object in the database cannot be filled in with user information because it has never been filled in. Therefore, in this case we should do one of the two:

- redirect such a user to `Information` and then `Interests` so that he would be forced to fill in initial information.
or
- redirect such a user to `UpdateMyProfile` where he would be forced to fill in initial information.

both of these ways would require to define some prop for a screen which would indicate whether a user is coming from the google sign in or not (because the layout would be different, the user in that case should not be able to go back, should not be able to quit and should not be re-directed to `Authentication` after `Interests`). Therefore, we need to define a flow and a mockup in Figma to illustrate how it needs to be implemented.

# How I did it
- Used firestore querying functions
- Added a new package for `Toast` messages

# How to verify it
Try out the registration flow yourself

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/79483254/93611de6-2b80-4107-81aa-53855cdab293

image

# Pre-merge checklist
- [ ] work correctly
- [ ] are fully tested


types/User.ts
gaspardthrl, line 1: I think using a class instead of a type would give us more flexibility...

What do you think? (I was thinking something similar to the Contact class I created in my PR)

gaspardthrl (approved):
As I told you by message, I really like the work you did! I was not sure about types vs classes at first but you explained clearly why it made more sense to use types for users. I don't have anything to say except that your code looks good to me!

GaelCondeLosada (approved):
Great work ! I didn't have time to check all files in detail but I tested it on android and everything works correctly (I checked the android box). Looks good to me !

GaelCondeLosada:
Don't forget to address/acknowledge sonar cloud issues before merging.

PR #133, opened 2024-04-29
feat: add contact by qr

GaelCondeLosada:
# What I did
- Finished the qr code scanning logic and implemented a rudimentary "add a contact" screen. 
- We cannot add a contact yet because the database does not keep track of contacts. 
- When scanning from inside the app, it is checked that the user exists and is not the same user as the one scanning, however when scanning from outside the app these cases aren't detected. I'm thinking of implementing these checks in another PR.

# How I did it
- Modified the QrScanScreen to parse the parsed link and act depending on the link.
- Added the dependency `use-debounce` and used it to make a debouncer for qr code scanning.
- Implemented the AddContactScreen.
- Had to modify a few tests to adapt for the changes.

# How to verify it
Try to scan this qr code with the scanner of the app and then with the scanner of your phone. You can also try to scan your own qr code, the app should tell you it's yourself.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/5f63ee16-53e0-4a1c-ae9f-cf9501f195f5

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


screens/QrScan/QrScanScreen.tsx
```diff
@@ -21,20 +30,32 @@ const QrScanScreen = () => {
...
-          },
-        ],
-        { cancelable: false }
-      )
+      const path = data.replace(Linking.createURL("/"), "")
+      const [route, id] = path.split('/')
+      if(route === "contact") handleUser(id)
+      else if (route === "event") handleEvent(id)
+      else showErrorToast("Qr code not recognized")
+      setTimeout(() => setQrScanned(false), 1500) // re-allow to scan QR in 1.5 second
```
AlbertoCentonze, line 58: Might be worth using a debouncer here: https://usehooks.com/usedebounce

GaelCondeLosada: What about this useDebounce: https://www.npmjs.com/package/use-debounce ? I'm asking you because I think you'll implement the debouncer for the search bar so it would be good if we use the same.

AlbertoCentonze: Any will do, we could even build the hook from scratch, it's a very simple logic! I'll follow whichever you pick. Yours seems better!

GaelCondeLosada: Ok I'll take this one then: https://www.npmjs.com/package/use-debounce

__test__/screens/QrScan/QrScanScreen.test.tsx
```diff
@@ -1,23 +1,70 @@
...
+  addListener: jest.fn(),
+  removeListener: jest.fn(),
+  reset: jest.fn(),
+  setParams: jest.fn(),
+  dispatch: jest.fn(),
+  isFocused: jest.fn(),
+  canGoBack: jest.fn(),
+  dangerouslyGetParent: jest.fn(),
+  dangerouslyGetState: jest.fn()
+} as unknown as NavigationProp<ParamListBase>
```
alexandre-mrt, line 21: Try not to use as unknown and put the correct type

alexandre-mrt (approved):
LGTM

AlbertoCentonze (approved):

PR #139, opened 2024-05-01
feat: interests from the database

Fyuryy:
# How I did it
I've written by hand all interests in the database. I decided to create an Interests object, containing a title, a category and an image URL, which will or not be used to import images.
Then I fetch the interests asynchronously. Therefore I had to change the functions logic so it would work with this new implementation. 
I tested the interest object and the Interest Screen

# How to verify it
Sign-in in the application and go to InterestsScreen, right after InformationScreen. You should see new interests added in the list

# Demo video
No need for video as the functionalities of the screen remain strictly the same. The only difference is the listed interests. For now, we have 50 interests of around 6 categories: Music, Sports, Games, General, Fitness, Cooking

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


GaelCondeLosada (changes requested):
Good PR, not much to say except one comment I've left. Everything works on my android device (I've checked the box).

GaelCondeLosada:
Also don't forget to address sonar cloud issues before merging.

GaelCondeLosada (approved):
LGTM

PR #141, opened 2024-05-01
Feat: events collection is fetched from database and events can be created by a user

GustaveCharles:
# What I did
I implemented multiple features:
- Events that appear in HomeScreen are fetched from database
- Events in the map are passed as props from the HomeScreen
- new Events can be created (watch demo to see where you can create them), are added in the database, and then fetched onto the HomeScreen and MapScree.

Note: there are still some issues in this feature. This is a list of notable problems:
- when an event is added, you need to refresh the app manually to see it in the HomeScreen. Ideally you'd like to swipe up to refresh the app.
- some values of the added event are hard coded (description, date, position,location,imageUrl)
- the fetching of the images from the imageUrl attribute does not seem to work
- We should also add a field to store the uid of the user that created the event, so that we can retrieve it in the ProfileScreen.

# How I did it
- Used firestore querying functions
- Used props between HomeScreen and MapScreen

# How to verify it

You can follow the same steps as mine on Expo Go. Be aware that for the new event to show, you need to reload the app manually.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/91184289/49d59256-5487-4dc4-b2af-4d0afd1b944d

# Pre-merge checklist
- [ ] work correctly
- [ ] are fully tested


__test__/firebase/ManageEvents.test.ts
```diff
@@ -0,0 +1,141 @@
...
+//     const events = await getAllPastEvents()
+//     expect(events.length).toBe(1)
+//     expect(events[0].title).toEqual("Past Event")
+//   })
+
+//   it("should fetch future events successfully", async () => {
+//     const events = await getAllFutureEvents()
+//     expect(events.length).toBe(1)
+//     expect(events[0].title).toEqual("Future Event")
+//   })
```
alexandre-mrt, line 139: if the test is not used remove it for better readability

__test__/screens/Home/HomeScreen.test.tsx
```diff
@@ -37,32 +45,32 @@ describe('HomeScreen', () => {
...
+    //       </SafeAreaProvider>
+    //     )
+
+    //     fireEvent.changeText(getByPlaceholderText('Search...'), 'Balelek')
+    //     expect(getByText('Balelek 2023')).toBeTruthy()
+    //     expect(queryByText('Event 2')).toBeNull()
+    //     expect(queryByText('')).toBeNull()
+    //     expect(queryByText('Abc')).toBeNull()
+
+    //     fireEvent.changeText(getByPlaceholderText('Search...'), '')
```
alexandre-mrt, line 61: Is it because you take into account that I will modifiy it later on, on my PR ? If yes it's ok to let that. Otherwise try to avoid

alexandre-mrt (changes requested):
LGTM. only answer about the comments or delete them

screens/Home/HomeScreen.tsx
```diff
@@ -11,153 +11,98 @@ import { useSafeAreaInsets } from "react-native-safe-area-context"
...
+            placeholder="Search..."
+            style={[styles.input, globalStyles.text]}
+            onChangeText={handleSearch}
+          />
         </View>
+        {/* Need to style the opacity to render it smaller*/}
+        <TouchableOpacity
+          style={styles.map}
+          onPress={() =>
+            navigation.navigate("EventMap",  {events: futureEvents})
```
GaelCondeLosada, line 80: I know you didn't introduce the error but if you can fix it...?

screens/Home/HomeScreen.tsx
```diff
@@ -11,153 +11,98 @@ import { useSafeAreaInsets } from "react-native-safe-area-context"
...
+
+      <View style={styles.containerEvent}>
+        <Text style={styles.title}>Future Events</Text>
+        <ScrollView style={styles.eventList}>
+          { }
+          {
+            renderEvents(true)
+          }
+        </ScrollView>
+        <Text style={styles.title}>Past Events</Text>
```
GaelCondeLosada, line 95: Try to use custom fonts from globalStyles for text.

screens/EventCreation/EventCreationScreen.tsx
```diff
@@ -12,10 +12,24 @@ import { useNavigation } from "@react-navigation/native"
...
+  const [location] = useState("Besac City")
+  // const [startDate, setStartDate] = useState<Date>()
+  // const [endDate, setEndDate] = useState("")
+  // const [imageUrl, setImageUrl] = useState("")
+//  const [point, setPoint] = useState({ x: 47.238458, y: 5.984155 })
+
+
+  const newEvent = async () => {
+    console.log("Creating event")
+    await createEvent("uid2", title, description, new Date(2025, 0, 1), { x: 47.238458, y: 5.984155 }, location, "imageUrl")
```
GaelCondeLosada, line 31: You could add a navigation back here I think, when the event is successfully created.

screens/Home/HomeScreen.tsx
```diff
@@ -11,153 +11,98 @@ import { useSafeAreaInsets } from "react-native-safe-area-context"
...
+  const [ setSearchQuery] = React.useState("")
 
-    // Further filter by search query if there is one
-    return eventsFilteredByDate.filter((event) =>
-      event.title.toLowerCase().includes(searchQuery.toLowerCase())
-    )
+  const handleSearch = (search: string) => {
+    setSearchQuery(search)
   }
+  useEffect(() => {
```
GaelCondeLosada, line 32: I believe you can use useFocusEffect() here https://reactnavigation.org/docs/use-focus-effect/ What it would do is reload the events every time this screen comes into focus. On the other hand the operation can be expensive so I'm just saying in case you didn't know. The decision to use it or not comes to you. You could also imagine to asynchronously check if there are new events on focus and update if necessary but not reload all events every time...

screens/Home/HomeScreen.tsx
```diff
@@ -11,153 +11,98 @@ import { useSafeAreaInsets } from "react-native-safe-area-context"
...
 
-  const handleSearch = (search: string) => {
-    setSearchQuery(search)
+    loadEvents()
+  }, [])
+
+  const renderEvents = (isFuture: boolean) => {
+    const events: Event[] = isFuture ? futureEvents : pastEvents
+    return events.map((event: Event) => (
+      <TouchableOpacity key={event.title}>
```
GaelCondeLosada, line 52: Here I think you should use the event uid as the key because it can be a problem when two events have the same name and thus the same key.

GaelCondeLosada (changes requested):
Few things to discuss/change but that's a good PR. Everything worked on my android device (I checked the box).

screens/Home/HomeScreen.tsx
GustaveCharles: Which error?

screens/Home/HomeScreen.tsx
GustaveCharles: I'll let the design guys modify that one, I think I did not change anything related to design, so that should not be my error.

screens/EventCreation/EventCreationScreen.tsx
GustaveCharles: Good point. I'll let @AlbertoCentonze , who is designing the screen do it.

screens/Home/HomeScreen.tsx
GustaveCharles: I will definetly look into it

GaelCondeLosada (approved):
LGTM

alexandre-mrt (approved):

Aidasvenc:
> Be aware that for the new event to show, you need to reload the app manually.

The events will need to be displayed in a `FlatList` where a pull-to-refresh is easy to set up. Opening an issue for that.

PR #149, opened 2024-05-05
feat: view event and view announcement screens

GaelCondeLosada:
# What I did
- I made a screen to see the details of an event. Note that the event data is dummy data for now because there is no function to fetch one event from the database.
- I made a screen to see the details of an announcement. Here we have the real announcement that is just passed as a prop.
- I made a small component for profiles pictures with a default icon in case we don't have access to the profile picture because I was always writing the same code for profile pictures...

# How I did it
- Made a simple component `ProfilePicture` that contains the code we always write for a profile picture. We should try to use this component for now on.
- Made a test for the `ProfilePicture` component.
- Made a screen `ViewEventScreen` in which I implemented the screen. The majority of the styles are in a global `ViewDetailsStyles.ts` file because the `ViewAnnouncement` uses the exact same layout and thus the exact same styles.
- Made a screen `ViewAnnouncement` which uses principally the `ViewDetailsStyles.ts`
- Added the two new screens to the main stack navigator.
- In order to link the navigation to these two screens I added a prop function (onPress) in `EventScreen`  and `AnnouncementScreen`. I also added the navigator in the `HomeScreen` and passed the functions needed to the announcement and event screens.
- I had to modify the tests for `EventScreen`, `AnnouncementScreen` and `HomeScreen` because I changed their props.

# How to verify it
You can launch the app and try to click on an event and an announcement. You can also run the tests.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/fd8cfc39-dc76-47ae-9427-7611b50031e6

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


__test__/screens/ViewDetails/ViewAnnouncementScreen/ViewAnnouncementScreen.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+// Create a dummy Announcement
+const dummyAnnouncement: Announcement = {
+    uid: "1",
+    title: "Community Art Exhibition",
+    location: "Downtown Community Center",
+    point: dummyPoint,
+    description: "Join us for an evening of art and conversation at our annual community art exhibition. Featuring works from local artists in a variety of mediums.",
+    interests: ["art", "community", "local culture"],
+    date: "2024-10-03T18:00:00Z"
+}
```
GustaveCharles, line 22: these dummy announcements/events should be taken out once and for all. It is not possible anymore to do this. If you want to mock the fetching of announcements, create a ManageAnnouncement in the firebase folder and mock a function there please.

screens/Home/AnnouncementScreen/AnnouncementScreen.tsx
```diff
@@ -71,7 +75,8 @@ const AnnouncementScreen = () => {
   }
 
   const renderItem = ({ item }: SectionListRenderItemInfo<Announcement>) => (
-    <TouchableOpacity >
+    <TouchableOpacity
+      onPress={() => {onAnnoucmentPress(item)}}>
       <AnnouncementCard {...item} />
     </TouchableOpacity>
   )
```
GustaveCharles, line 82: Same in this screen overall. There cannot be mocking values everywhere and then hope for the best for the backend guys.

screens/Home/HomeScreen.tsx
```diff
@@ -9,15 +9,15 @@ import React, { useState } from "react"
...
 import EventScreen from "./EventScreen/EventScreen"
 import AnnouncementScreen from "./AnnouncementScreen/AnnouncementScreen"
+import { NavigationProp, ParamListBase } from "@react-navigation/native"
 
+interface HomeScreenProps {
+  navigation: NavigationProp<ParamListBase>
+}
 
-const HomeScreen = () => {
+const HomeScreen = ({ navigation }: HomeScreenProps) => {
```
GustaveCharles, line 18: why do you wrap a props in an interface? I think that you do not need to do that. It was the problem that @alexandre-mrt had.

screens/ViewDetails/ViewAnnouncementScreen/ViewAnnouncementScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+type RootStackParamList = {
+  ViewAnnouncement: {
+      announcement: Announcement;
+  }
+}
+
+type ViewAnnoucementScreenRouteProps = RouteProp<RootStackParamList, "ViewAnnouncement">
+
+const ViewAnnoucementScreen = () => {
+  const { announcement } = useRoute<ViewAnnoucementScreenRouteProps>().params
```
GustaveCharles, line 18: I may be wrong, but same here, I do not think we need all of that for passing an Announcement as prop. I'd like to know your take on this.

screens/ViewDetails/ViewEventScreen/ViewEventScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+    ViewEvent: {
+        // Here we're getting a uid and not an event because it is possible that the app opens on this screen (by QR)
+        uid: string;
+    }
+}
+  
+type ViewEventScreenRouteProps = RouteProp<RootStackParamList, "ViewEvent">
+
+const ViewEventScreen = () => {
+  const { uid } = useRoute<ViewEventScreenRouteProps>().params
```
GustaveCharles, line 37: Sorry for bothering that much, but I really think we should start doing things another way. Thus, in this example, I would suggest that you go to the ./firebase/ManageEvents.tsx file, and create a function FetchSingleEvent, in which you only return this dummy event.

GustaveCharles, line 37: I also think again, that you do not need a root stackParamList, but just the ViewEventScreenRouteProps. You can maybe take a look at how in #141 the event was passed as a props to the MapScreen. Again, I may be mistaken, so tell me your way of thinking please.

GustaveCharles (changes requested):
Some structural changes need to be addressed so that everyone can do his job in the best and easiest way possible. Thanks in advance for the changes!

__test__/screens/ViewDetails/ViewAnnouncementScreen/ViewAnnouncementScreen.test.tsx
Gaellosada: I don't really see the point in mocking the database. If you look at the code the announcement is just passed as props to the screen. I would need to mock the database then fetch the announcement and them pass it at prop why not just pass it right away ? There is not database implicated in this screen so I'm only testing the UI.

GaelCondeLosada: I don't really see the point in mocking the database. If you look at the code, the announcement is just passed as props to the screen. I would need to mock the database then fetch the announcement and them pass it at prop, why not just pass it right away ? There is not database implicated in this screen so I'm only testing the UI.

screens/Home/AnnouncementScreen/AnnouncementScreen.tsx
GaelCondeLosada: I'm not doing this screen, I just had to modify it to allow the navigation to work. But don't worry we can do a PR to clean that when we have the database fetching function for announcements.

screens/Home/HomeScreen.tsx
GaelCondeLosada: I did it because when navigating with parameters, typescript gives an error without it (when calling navigation.navigate(...)). I didn't want to leave typescript errors.

screens/ViewDetails/ViewAnnouncementScreen/ViewAnnouncementScreen.tsx
GaelCondeLosada: Same as above, it's to make typescript happy. You can try to go on this branch and do without that, you'll understand what I'm saying. There is maybe a simpler way to do it but I didn't manage to find one. The thing is you have to "register" the parameter in the navigation so the parameters can be "mapped" and you can get them back at arrival (plus you have to specify the type for typescript). If you find another way to do it tell me and we can change it.

screens/ViewDetails/ViewEventScreen/ViewEventScreen.tsx
GaelCondeLosada: I could do it but I think it's better that I don't modify too much files to not have conflicts. Especially if this function has to be implemented then if I make the function the one implementing it will have to wait for my PR to be merged... this is not really efficient. I think it's better to develop locally and then link everything in a later PR when everything is ready and we won't have conflicts. If you disagree I can change it and add a function.

GaelCondeLosada: It's the same reason as explained above. The implementation in #141 way actually fixed in a later PR in order to remove the typescript error. You can look at the code of the MapScreen in main, you'll find the same implementation.

screens/ViewDetails/ViewAnnouncementScreen/ViewAnnouncementScreen.tsx
GustaveCharles: Thanks for the info. If it works, there is no worry to have I guess. I learnt swomething new today.🚀

GustaveCharles:
Hey Gael, thanks a lot for the clarifications. I get a problem when testing on my android emulator. Basically, the problem is that when I scroll, I presses automatically on an event, without me pressing. I attach a video, that hopefully showcases it:



https://github.com/uniconnect-epfl/uniconnect/assets/91184289/99fae857-6729-440b-b4a0-a0e4f660736d

GaelCondeLosada:
> Hey Gael, thanks a lot for the clarifications. I get a problem when testing on my android emulator. Basically, the problem is that when I scroll, I presses automatically on an event, without me pressing. I attach a video, that hopefully showcases it:
> Enregistrement.de.l.ecran.2024-05-09.a.11.36.55.1.1.mov

Hey I made I fix, it works on my device. Could you test again ?

GustaveCharles (approved):
Great PR Gael. Thanks a lot!

PR #151, opened 2024-05-05
feat: user qr scanning finalization

GaelCondeLosada:
# What I did
- Redirected to the `ExternalProfileScreen` when scanning someone's QR code.
- Made it so `ExternalProfileScreen` adapts if we are friend or not with the external profile (right now hard coded but will adapt to the database later). If we are not friends the screen will have a button "Add", if we are already friends we will have a button "Message" and "Remove"

# How I did it
- Deleted the `AddContactScreen` file and tests.
- Redirected the navigation and the deep linking to the `ExternalProfileScreen`.
- Modified the `ExternalProfileScreen` to fetch the user data and the external profile from the database and adapt to if the user and the external are friends or not. The adding and removal of friends will be implemented when the database can support if.

# How to verify it
You can scroll the contact list and click on contacts to see details. You can also try to scan a QR code generated with your expo go session (from inside or outside the app). You can of course also run the tests. If you want to test when you aren't friend with someone you can just modify the line 65 of `ExternalProfileScreen` and set isFriend to false.

I looked a bit into why from a version to another the QR weren't working and here is the reason. This means you need the QR to be generated into the same expo go session you are scanning from. We of course won't have this problem in production build.

Capture

You can read more at [this link](https://docs.expo.dev/guides/linking/)

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/8c8bee91-8b89-4e2b-9e84-6692ac059324

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


Aidasvenc:
Hey, I think you have a wrong demo.

screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
```diff
@@ -1,55 +1,77 @@
...
-};
+  }
+}
 
-type ExternalProfileScreenRouteProp = RouteProp<RootStackParamList, 'ExternalProfile'>;
-*/
+type ExternalProfileScreenRouteProp = RouteProp<RootStackParamList, "ExternalProfile">
 
 const ExternalProfileScreen = () => {
+  const { uid } = useRoute<ExternalProfileScreenRouteProp>().params
```
Aidasvenc, line 27: Should rename the variable to something like externalUserId, because it is a bit confusing with the 2 uids being present in this screen.

Aidasvenc (commented):
only a minor change in variable names is needed and you will be good to go!

Fyuryy (approved):
LGTM

GaelCondeLosada:
> Hey, I think you have a wrong demo.

You're right that's my bad! I changed it.

screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
```diff
@@ -59,29 +81,41 @@ const ExternalProfileScreen = () => {
...
+            <View style={profileStyles.buttonsContainer}>
+              <TouchableOpacity 
+                style={[profileStyles.button, styles.uniqueButton]}
+                onPress={() => alert("To come")}>
+                <View style={profileStyles.horizontalContainer}>
+                  <Text style={[globalStyles.boldText, profileStyles.buttonText]}>Add</Text>
+                </View>
+              </TouchableOpacity>
+            </View>
+          )}
```
alexandre-mrt, line 114: You could put those in two separate function that could improve readability.

alexandre-mrt (approved):
Appart from little improvements possible in readability. LGTM

PR #163, opened 2024-05-05
fix: homescreen layout

alexandre-mrt:
# What I did
I implemented the columns layout for the home screen
Resolved little issues with Homescreen layout on android.
Solved a problem from the type event that prevented the map to work correctly

# How I did it
I had to group the events by two to render it with section List. Section list did not have the column option so I managed to display the events 2 by 2.
I fixed the test for the Homescreen
I replaced the previous attribute latitude and longitude by x and y to allow correct render on the map.
# How to verify it
You can see the columns layout directly on the Homescreen
Test the map that renders without errors now 

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/108724823/13f56131-4d8d-425d-a8b6-4cbc451f547d

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


components/EventCard/styles.ts
```diff
@@ -1,28 +1,38 @@
...
+    borderColor: peach,
     borderRadius: 10,
-    elevation: 5,
-    margin: 10,
-    overflow: 'hidden',
+    borderWidth: 2,
+    flex: 1,
+    height: 202,
+    margin: 5,
+    overflow: 'scroll',
```
Aidasvenc, line 14: should not use this, instead use `hidden` to cover overflowing layout.
```suggestion
    overflow: 'hidden',
```

components/EventCard/styles.ts
```diff
@@ -1,28 +1,38 @@
...
+    
     shadowColor: shadowColor,
     shadowOffset: {
       height: 2,
       width: 0,
     },
     shadowOpacity: 0.25,
     shadowRadius: 3.84,
+    width: 172,
```
Aidasvenc, line 23: the width should be set to percentage terms, never use constant width parameters. E.g. here we want it to be `95%` of the whole screen width

components/EventCard/styles.ts
```diff
@@ -1,28 +1,38 @@
...
+    
   },
   description: {
+    
     color: lightGray,
     fontSize: 14,
+    fontStyle: 'italic',
+    textAlign: 'right',
```
Aidasvenc, line 31: You seem to also forget to use our own custom font. Don't forget to define it with a `fontFamily` parameter

screens/Home/EventScreen/EventScreen.tsx
```diff
@@ -52,32 +77,54 @@ const EventScreen = () => {
...
+          <EventCard {...event} />
+        </TouchableOpacity>
+      ))}
     </View>
   )
+  
+
+  const renderSectionHeader = (info : {section: SectionListData<Event[], DefaultSectionT>} ) => {
+    return (
+      <View style={{ backgroundColor: defautlBackgroundColor}} >
```
Fyuryy, line 114: here it should be default and not defautl

screens/Home/EventScreen/EventScreen.tsx
```diff
@@ -52,32 +77,54 @@ const EventScreen = () => {
...
+          disabled={event.uid === "dummy"}
+        >
+          <EventCard {...event} />
+        </TouchableOpacity>
+      ))}
     </View>
   )
+  
+
+  const renderSectionHeader = (info : {section: SectionListData<Event[], DefaultSectionT>} ) => {
```
Fyuryy, line 112: I feel like this way of writing is quite confusing... You could declare an interface like so: 

interface SectionHeaderProps {
  section: SectionListData<Event[], DefaultSectionT>;
} 

and when declaring the const do it like this: 
const RenderSectionHeader: React.FC <.SectionHeaderProps >  = ({ section }) => ....

screens/Home/stylesScreen.tsx
```diff
@@ -45,28 +48,45 @@ export const styles = StyleSheet.create({
...
+
+  },
   map: {
     
-    alignItems: 'center',
     backgroundColor: lightPeach,
     borderColor: peach,
     borderRadius: 40,
     borderWidth: 2,
     fontWeight: 'bold',
```
Fyuryy, line 69: You can use the global styles for the fonts

screens/Home/stylesScreen.tsx
```diff
@@ -79,14 +99,19 @@ export const styles = StyleSheet.create({
...
     marginLeft: "3%",
     marginRight: "50%",
-    
+   
    
   },
   title: {
     color: black,
     fontSize: 20,
     fontWeight: 'bold',
```
Fyuryy, line 108: same here, you can use the global font styling

screens/Maps/EventMap.tsx
```diff
@@ -1,10 +1,11 @@
 import MapView, { Callout, Marker, PROVIDER_GOOGLE } from "react-native-maps"
-import React from "react"
+
 import { RouteProp, useNavigation, useRoute } from "@react-navigation/native"
 import { View, Text, TouchableOpacity } from "react-native"
 import styles from "./styles" // Import styles
 import { Ionicons } from "@expo/vector-icons"
 import { Event } from "../../types/Event"
+import React = require("react")
```
Fyuryy, line 8: Why do you import react this way? does it differ from the usual import? It always break my code so if it does work for you I recommend you changing it before pushing

Fyuryy (changes requested):

screens/Maps/EventMap.tsx
AlbertoCentonze: This should not even be allowed by the linter...

AlbertoCentonze:
Please internal padding of the cards for the event should not be this big!

screens/Home/EventScreen/EventScreen.tsx
```diff
@@ -52,32 +77,54 @@ const EventScreen = () => {
...
+    }
+    return grouped
+  }
+  
+  
+
+
+  const renderItem = ({ item }: SectionListRenderItemInfo<Event[]>) => (
+    <View style={styles.row}>
+      {item.map((event) => (
```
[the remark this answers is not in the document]
alexandre-mrt: It's a dedicated function for SectionList. It would be pointless to add a Flat list inside a Section List. But in a general manner I agree

Fyuryy:
Also, I just retested it on an android emulator and when you leave the event screen and go back to it afterwards, all the events disappear. This might be because you don't have the fetching feature yet, but it's still worth taking a look.

screens/Home/EventScreen/EventScreen.tsx
Fyuryy: Ok

alexandre-mrt:
> Also, I just retested it on an android emulator and when you leave the event screen and go back to it afterwards, all the events disappear. This might be because you don't have the fetching feature yet, but it's still worth taking a look.

The state persistence as been implemented in another pr

Fyuryy (approved):

PR #165, opened 2024-05-06
feat: refine user flows & styling

Aidasvenc:
# What I did
Addressed the following issues and user feedback mentioned in #144:

- Make section tabs easier to press on.
- Improve the password: make an option to hide the password with "****" in registration and finish all checks (some of the checks like password strength are not functional I don't remember which ones).
- After adding a description, the fact that there is still a button “Add a description” made users feel weird. Change the button to “Modify description” or remove it after the user has put their description.
-  On login the button "Continue with google" should have a capital G.
- In registration, “Please check your data” is not a good error message, it doesn’t tell the user what’s wrong.
- Make the screen to view all the announcements in a list.

Also, did some styling changes in order to have a more consistent design language and to focus on visibility of main components in the application.

# How I did it
- Changed layouts of different components and screens.
- Changed styling of different components and screens

# How to verify it
- Run the app and explore it.

# Demos
WhatsApp Image 2024-05-06 at 22 26 23_19fa951f
WhatsApp Image 2024-05-06 at 22 26 23_ee5ddf31
WhatsApp Image 2024-05-06 at 22 26 24_d4da9f1a
WhatsApp Image 2024-05-06 at 22 26 24_442e5819

# Pre-merge checklist
The changes I have introduced:
- [x] work correctly
- [x] do not break other functionalities
- [ ] work correctly on Android
- [x] are fully tested

PR #166, opened 2024-05-07
fix: android layout bugs

Fyuryy:
# What I did
 - Fixed android related bugs such as the position of the lowbar whenever the keyboard is opened. Now the lowbar is well fixed throughout all the registration workflow. Moreover, when the keyboard is opened, it hides correctly the screen without pushing it and the lowbar is hidden.
 - Limited the date of birth to the date of the actual day
 - Passwords now do not appear in plain text but with <b>*</b>'s

# How I did it
Used useEffect and listeners to hide the keyboard and changed the zIndex of some components so they are hidden by the keyboard when it's opened.  

# How to verify it
By going through the registration flow like an user would, it's possible to check all the changes I made. 

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/screens/Onboarding/styles.ts
+++ b/screens/Onboarding/styles.ts
@@ -15,11 +15,6 @@ import {
 } from "../../assets/global/constants"
 
 const styles = StyleSheet.create({
-  border: {
-    flex: 1,
-    flexDirection: "column",
-    justifyContent: "space-between",
-  },
   button: {
     alignItems: "center",
     backgroundColor: lightPeach,
@@ -37,7 +32,6 @@ const styles = StyleSheet.create({
     width: "100%",
   },
   buttonGoogle: {
-    alignItems: "center",
     backgroundColor: white,
     borderRadius: BUTTON_RADIUS,
     borderWidth: 1,
@@ -53,36 +47,25 @@ const styles = StyleSheet.create({
     width: "100%",
   },
 
-  buttonPlaceholder: {
-    alignItems: "center",
-    flex: 1,
-    flexDirection: "row",
-    justifyContent: "flex-start",
-  },
   buttonText: {
     color: black,
     fontSize: BUTTON_TEXT_FONT_SIZE,
   },
-  buttonTextGoogle: {
-    color: black,
-    fontSize: BUTTON_TEXT_FONT_SIZE,
-    fontWeight: "bold",
-  },
+
   container: {
     alignItems: "center",
     flex: 1,
     gap: 1,
     padding: 20,
   },
+
   footer: {
-    display: "flex",
+    flex: 1,
     flexDirection: "row",
     position: "absolute",
     zIndex: -1,
   },
-  icon: {
-    paddingRight: 25,
-  },
+
   image: {
     height: 250,
     marginBottom: 10,
--- a/screens/Onboarding/styles.ts
+++ b/screens/Onboarding/styles.ts
@@ -1,4 +1,4 @@
-import { StyleSheet } from 'react-native' 
+import { StyleSheet } from "react-native"
 import {
   white,
   black,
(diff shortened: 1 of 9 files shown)


screens/Onboarding/styles.ts
```diff
@@ -1,88 +1,71 @@
...
+    width: "100%",
   },
   buttonGoogle: {
-    alignItems: 'center',
     backgroundColor: white,
     borderRadius: BUTTON_RADIUS,
     borderWidth: 1,
-    display: 'flex',
-    flexDirection: 'row',
+    display: "flex",
```
Aidasvenc, line 38: i am pretty sure that if you delete `display: "flex"`, the button is going to look the same way

screens/Onboarding/styles.ts
```diff
@@ -1,88 +1,71 @@
...
-  footer: {
-    display: 'flex',
-    flexDirection: 'row',
-    position: 'absolute',
 
+  footer: {
+    flex: 1,
+    flexDirection: "row",
+    position: "absolute",
+    zIndex: -1,
```
Aidasvenc, line 66: Why did you need a `zIndex: -1` here? Wasn't `bottom: insets.bottom` not enough for the footer to be anchored to the bottom?

screens/Registration/InformationScreen/InformationScreen.tsx
```diff
@@ -30,8 +30,32 @@ const InformationScreen: React.FC = () => {
...
+    })
+    const hideSubscription = Keyboard.addListener("keyboardDidHide", () => {
+      setKeyboardVisible(false)
+    })
+
+    return () => {
+      showSubscription.remove()
+      hideSubscription.remove()
+    }
+  }, [])
```
AlbertoCentonze, line 47: This seems to be recurring, why don't you create a custom hook called `useKeyboard` to access this more easily? I don't think you've ever made custom hook but it's really easy can help you out if needed!

screens/Registration/AuthenticationScreen/AuthenticationScreen.tsx
```diff
@@ -116,24 +135,32 @@ const AuthenticationScreen: React.FC = () => {
...
+          ></InputField>
+        </View>
 
-        <LowBar nextScreen="HomeTabs" buttonText="Confirm" authenticate={submitForm} />
+        <View style={[styles.footer, { bottom: insets.bottom }]}>
+          {!keyboardVisible && <LowBar
+            nextScreen="HomeTabs"
+            buttonText="Confirm"
+            authenticate={submitForm}
+          />}
```
Aidasvenc, line 162: Why do you want to hide the LowBar when keyboard is on?

AlbertoCentonze (changes requested):
Hey, nice job!

Since I've seen a lot of `absolute` used in the CSS I would recommend testing the layout on different screen sizes since absolute can easily create footguns in this kind of scenario. I think you can safely do this by playing around with the window size of a react native web instance, so that you don't have to install a bunch of simulators.

screens/Registration/AuthenticationScreen/AuthenticationScreen.tsx
Aidasvenc: Sorry for a duplicate solution, but take a look at my code in #165 for how i anchored the LowBar to the bottom. Ideally, you should not wrap the LowBar in a View each time, but rather change the styling for the LowBar itself. Take a look at my solution to see more.

screens/Registration/AuthenticationScreen/styles.ts
```diff
@@ -1,52 +1,61 @@
...
-    alignItems: 'center',
-    alignSelf: 'center',
+    alignItems: "center",
+    alignSelf: "center",
     backgroundColor: lightPeach,
     borderColor: peach,
     borderRadius: BUTTON_RADIUS,
     borderWidth: 3,
-    display: 'flex',
+    display: "flex",
```
Aidasvenc, line 13: Same here

Aidasvenc (commented):
Nice fixes! Inspect a couple of details regarding the styling and you will be good to go

screens/Onboarding/styles.ts
Fyuryy: The footer was anchored at the bottom without the zIndex: -1 indeed. The problem was when opening the keyboard, the footer came up. So I solved it this way. Do you think it would be better to do as with the other screens, with a useEffect?

screens/Registration/AuthenticationScreen/AuthenticationScreen.tsx
Fyuryy: The reason I want to hide it is because on IOS, the LowBar was totally fixed but on android the lowbar was fixed until the keyboard was opened and pushed it on top of it

screens/Onboarding/styles.ts
Fyuryy: Indeed

screens/Registration/AuthenticationScreen/styles.ts
Fyuryy: Indeed

screens/Registration/InformationScreen/InformationScreen.tsx
Fyuryy: Done!

AlbertoCentonze (approved):

PR #168, opened 2024-05-07
feat: announcements are fetched from database and can be added to it

GustaveCharles:
# What I did

# How I did it

# How to verify it

# Demo video

# Pre-merge checklist
- [ ] work correctly
- [ ] are fully tested


__test__/firebase/ManageAnnouncements.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  const mockGetDoc = jest.fn().mockImplementation((uid: string) =>{
+    if (uid === "0") {
+      return Promise.resolve({ exists: jest.fn(() => false)})
+    }
+    else if (uid === "1") {
+      return Promise.resolve({ exists: jest.fn(() => true)})
+    }
+    else {
+      return Promise.reject()
+    }
```
alexandre-mrt, line 20: Could you test the case where we try to create an uid similar to another. This should reject the creation, I think that's an important test as it caused a lot of problems before.

components/AnnoucementCard/AnnouncementCard.tsx
```diff
@@ -1,7 +1,7 @@
 
 import { styles } from "./styles"
-import React,{ View, Text }  from "react-native"
-import { Announcement } from "../../types/Annoucements"
+import React, { View, Text } from "react-native"
+import { Announcement } from "../../types/Annoucement"
```
alexandre-mrt, line 4: The file is named Annoucement. It appeared in a previous PR, if you could just change it's name to Announcement could be awesome !

Fyuryy:
Looks good to me! I've tested it on an android emulator and it works. However, I have the feeling that you and Alex might end up having a few conflicts as you are working on the same screens. Having checked both branched, I think it's important you both clarify on what exactly are you working!

firebase/ManageAnnouncements.ts
```diff
@@ -0,0 +1,46 @@
...
+            interests: interests,
+            date: date
+        })
+
+        showSuccessToast("Announcement created successfully!")
+    } catch (error) {
+        showErrorToast("There was an error storing your announcement data, please try again.")
+    }
+}
+
```
alexandre-mrt, line 24: Could be good to add a check preventing duplicate uid.

Fyuryy:
And some comments on the creation of an event/announcement. Whenever we create an event, it goes to the announcements tab. Moreover, we shouldn't be able to add events without date but I think this is a problem for the next sprint. Overall good job, your code is very clean.

Fyuryy (approved):
GJ

firebase/ManageEvents.ts
```diff
@@ -31,6 +30,7 @@ const formatEvent = (doc): Event => {
 
   const event = {
     ...data2,
+    uid: doc.id,
```
alexandre-mrt, line 33: this should be generated directly and we should not allow it to be passed as an argument. It's essential to ensure uniqueness of the uid

alexandre-mrt (commented):
Should handle the uniqueness of the uid. Otherwise LGTM

PR #172, opened 2024-05-09
feat(graph): simple mechanism to display friends of friends on the graph 

gaspardthrl:
# What I did
- I implemented the possibility to view friends of your friends on the graph with upgraded gestures
- I also made the animations smoother and improved the overall experience.
- I implemented double tap on nodes to display the modal and long press to select a node accompanied by vibrations to know when a node is selected.

# How I did it
- I used the Animated library as well as the Vibration library.

# How to verify it
1. Go to the graph section, you should see only 4 friends.
2. Double click on any node and you should see the modal appear (click anywhere outside of the modal to make it disappear)
Note that you should see both a zoom in and a zoom out animation when displaying the modal.
3. Press on any node without removing your finger until you feel a vibration. You should see the following behaviors depending on which node you clicked:
a. If you clicked on the node with an orange background at first, a toast error message should appear. In this case repeat step 3.
b. If you clicked on any other node, you should have witnessed a zoom in / zoom out animation and depending on which one you clicked on you might see new nodes (clicking on Bob Johnson or Carol Williams will result in additional nodes as they have friends while clicking on Henry Taylor or Isabella Rodriguez won't result in any new node). Moreover, you should see that the node at the center of the screen is the one you pressed.
If you don't see any new nodes long press on the node you just pressed which should reset the layout and repeat step 3.
4. Now that we have additional nodes, long press on any of them. You should see one of the following behaviors:
a. If you clicked on the node with the orange background or on the node you just clicked then you should find the original graph (the layout might have changed)
b. If you clicked on a node that was already here at the beginning of step 3., you should see the same behavior as in step 3.
c. If you clicked on one of the new node you should have seen a toast message telling you that you can only display friends of friends.
5. Changing screen and coming back should not change the graph (beside the zoom out when displaying the graph), that is if you displayed friends of a friend they should still be here if you did not reset the graph by pressing on the node with an orange background or on the node you pressed before.

# Demo video
https://github.com/uniconnect-epfl/uniconnect/assets/81919237/8e844eae-cff3-4e71-8194-ab153e44a5e1

# Additional Information
Note that part of the code for accessing the actual database is already written in this PR but we need to resolve problems with firebase before using it so I commented the code and used the mockup contacts.
Note also that the following (small) issue being that sometime the zoom is performed on a node not entirely in the displayed view and therefore appears as cut off, will be addressed in a later PR
Finally, note that retrieval by similarity and async storage for friends data will be implemented in a later PR (both are shown by TODOs in the code)

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


components/Graph/ForceDirectedGraph/ForceDirectedGraph.tsx
GaelCondeLosada, line 1: It's not a request for changes but just a remark. A file of close to 500 lines is hard to understand for me and could become harder to maintain in my opinion. You could try to split up the logic if you have time. (I know we don't have much time for refactors, so only if you have free time...)

firebase/User.ts
GaelCondeLosada, line 1: Hey try not to modify too much files if you don't need to. This is a call to get conflicts with others.

firebase/firebaseConfig.ts
GaelCondeLosada, line 1: Same thing here. If you want to refactor the code I think you should do a separated PR. If you don't have conflicts I think it's ok for this time.

GaelCondeLosada (commented):
Hey thanks for this PR ! I think this is good to be merged. I tested it on my android device and everything works but there are a few things:
- First the animation are now smooth that's good!
- Second on my device the pictures are quite pixelated when zooming.
- Third I cannot long click on the first click when arriving on the graph, if I do it, it just focuses on the node.

Of course the "issues" can by addressed in another PR in my opinion.

Here is a video to show it on my device:

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/8c6efb5e-912b-4a7b-b296-5cebcb4909ab

GaelCondeLosada (approved):
Forgot to Approve on the last review...

components/Graph/ForceDirectedGraph/ForceDirectedGraph.tsx
gaspardthrl: This will be addressed in my next PR

firebase/firebaseConfig.ts
gaspardthrl: oh I think it's when I saved the file it got refactored automatically

PR #184, opened 2024-05-13
fix: upgraded expo sdk and fix some small issues

GaelCondeLosada:
# What I did
- Upgraded expo SDK to 51
- Upgraded dependencies that the new SDK needed.
- Fixed issues with the layout of search bars.
- Fixed layout issues in the profile screen.
- Fixed issues with camera authorizations on some devices.
- Fixed a problem with the QR scanning debouncer.
- Fixed a problem with the navigation when scanning a QR code.

# How I did it
- Updated expo SDK and dependencies.
- I improved the `InputField` component and used it for all search bars so the layout is uniform and function everywhere in the app.
- I changed some of the registration layout because some of it got broken when I modified the InputField.
- Changed the styles in `MyProfile` and  `globalProfileStyles` to fix all layout issues in the `MyProfile`  screen.
- Added the camera authorizations as plugin.
- I also removed some unused styles.
- Modified a wrong navigation parameter after having scanned a QR code.

# How to verify it
Open the app in expo go and:
- Check that the search bars function correctly in `HomeScreen`, `ContactList`, `ContactGraph` and `ExternalProfileScreen`
- Go to `MyProfile` and check that the layout is good.
- Try to use the camera to scan someone's QR code (note that if the QR does not come from the same Expo go session, it will say "Qr not recognized" but if the camera scans successfully, it means it is working). **I need someone to test this for me becuase it always worked on my device.**

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/9836e978-1c2e-4d82-8294-ea911915e14b

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


screens/Contacts/ContactGraph/ContactGraph.tsx
```diff
@@ -59,8 +58,8 @@ const ContactGraph = ({
     // Dismiss the keyboard when the user taps outside of the search bar
     <TouchableWithoutFeedback onPress={() => Keyboard.dismiss()}>
       <View style={styles.container}>
-        <TextInput
-          style={[globalStyles.text, styles.searchBar]}
+        <InputField
+          label=""
```
Aidasvenc, line 62: why do you add an empty label? What purpose does it serve?

screens/Contacts/ContactList/ContactList.tsx
```diff
@@ -69,14 +69,16 @@ const ContactList = ({ onContactPress, contacts }: ContactListProps) => {
 
   return (
     <View style={styles.container}>
-      <TextInput
-        style={[globalStyles.text, styles.searchBar]}
+      <InputField
+        label=""
```
Aidasvenc, line 73: Same here

components/InputField/styles.ts
```diff
@@ -7,6 +7,7 @@ const styles = StyleSheet.create({
     borderRadius: 25,
     borderWidth: 1,
     height: 50,
```
Aidasvenc, line 9: the input should have a height of 40, because now it is taller than the button next to it in the home screen

components/InputField/styles.ts
```diff
@@ -21,20 +22,12 @@ const styles = StyleSheet.create({
   },
 
   section: {
-    display: "flex",
-    flexDirection: "column",
+    alignItems: "center",
+    flexDirection: "row",
+    flex: 1,
     justifyContent: "center",
+    maxHeight: 70,
```
Aidasvenc, line 29: Since the input's height is fixed, there should be no need to specify `maxHeight` for the section itself

screens/Profile/profileStyles.ts
```diff
@@ -36,12 +36,14 @@ export const profileStyles = StyleSheet.create({
...
     profileContainer: {
+        flex: 1,
+        marginTop: "35%",
         paddingHorizontal: "2.5%",
-        width: "100%"
     },
     topBackground: {
         backgroundColor: lightPeach,
         height: "20%",
+        position: "absolute",
```
Aidasvenc, line 46: Could you explain the reasoning why you added a position: "absolute"? What was the problem with the layout before?

screens/Contacts/ContactGraph/ContactGraph.tsx
GaelCondeLosada: The InputField needed it but I made the label prop optional, that's definitely better.

components/InputField/styles.ts
GaelCondeLosada: It's because I need the flex: 1 parameter so the bar fills the maximum amount of space horizontally, but I don't want it to take space vertically. That's the way I found to do it, do you think I should do it another way ?

screens/Profile/profileStyles.ts
GaelCondeLosada: Sadly the database is not working on my end for now so I'll show you on the figma. This is in order to get this overlap of the profile picture over the top background.

Capture

components/InputField/styles.ts
Aidasvenc: oh, i see. I would use the `width: "100%"` for such a component, but using flex + bounded height works as well!

screens/Profile/profileStyles.ts
Aidasvenc: I see. Nice approach!

Aidasvenc (approved):
Thank you for addressing the details, it looks good now!

gaspardthrl:
LGTM

PR #187, opened 2024-05-13
ci: now build is triggered only on release

AlbertoCentonze:
# What I did
Now continuous integrations only build on release or on dispatch workflow. This change was necessary as we were regularly running out of free builds.

# How to verify it
Click on the button to trigger a manual workflow. Try to do a release (and delete it after please).


GustaveCharles (approved):
LGTM

PR #188, opened 2024-05-13
feat: choose point on map for events

GaelCondeLosada:
# What I did
- Added a screen for the user to select a location on the map.
- Added a button that redirect to this screen in `EventCreationScreen`

# How I did it
- Created the `SelectLocationScreen` and added it to navigation. The screen manages it's own point state internally (meaning the currently selected location is managed internally) but when navigating you can give a call back function for when the date change, the same way as a TextInput.
- Used `react-native-maps` to implement the map.
- Added a button in `EventCreationScreen` that redirects to this screen with the right callback function.
- Modified tests to reflect changes.

# How to verify it
Launch the application in expo go:
- Go into `EventCreationScreen`
- Create a new event and add a location (Note that there is a warning but it will be fixed in another PR also fixing the warnings in the other maps)
- Try to click at multiples places (note that if you click on the name of a city for example the click does not register)
- Try to confirmate the location
- Go back to modify the location, see if it registered correctly
- Save the event
- You can go in the database and check if the location point is good

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/afc0c2d8-a98b-443e-ab70-1c95c2b3e9a9

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/screens/EventCreation/EventCreationScreen.tsx
+++ b/screens/EventCreation/EventCreationScreen.tsx
@@ -47,14 +47,10 @@ const EventCreationScreen = ({ navigation, isAnnouncement }: EventCreationScreen
     console.log("Date:", date.toDateString())
     console.log("Interests:", interests)
 
-    if(point === undefined){
-      showErrorToast("An event needs a location!")
+    if (isAnnouncement) {
+      newAnnouncement()
     } else {
-      if (isAnnouncement) {
-        newAnnouncement()
-      } else {
-        newEvent()
-      }
+      newEvent()
     }
 
     // after the user has filled out the form
@@ -146,7 +142,7 @@ const EventCreationScreen = ({ navigation, isAnnouncement }: EventCreationScreen
               value={location}
               onChangeText={setLocation}
             />
-            
+
             <Pressable
               style={styles.section}
               onPress={() => {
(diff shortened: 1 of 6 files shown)


screens/EventCreation/EventCreationScreen.tsx
```diff
@@ -142,6 +136,17 @@ const EventCreationScreen = ({ isAnnouncement }: EventCreationScreenProps) => {
           </View>
         )}
         <View style={styles.bottomButtons}>
+          <Pressable style={styles.buttonBase}>
+            <Text
+              onPress={() => {
+                console.log("going to select a location")
```
alexandre-mrt, line 142: If it's working fine you can remove it.

screens/SelectLocation/SelectLocationScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+            onPress={() => {
+                navigation.goBack()
+                console.log("user clicked on confirmate")
+                console.log("confirmating location")
+                console.log("we ve confirmated the location")
+                console.log("the location was entered we are now returning")
+                console.log("is location defined ?")
+                console.log("we don't know yet")
+                console.log("we can now return")
+                console.log("going back to precedent screen")
```
alexandre-mrt, line 76: Delete if everything is working please. The console can be un-readable otherwise.

alexandre-mrt (changes requested):
Just a reminder to delete the log. Otherwise LGTM

alexandre-mrt (approved):
LGTM

AlbertoCentonze (commented):
Hey Gael, thanks for the PR! Can you help me understand why you passed navigation as a prop instead of using the hook as we were doing before?

Also I saw managed to pass functions through `navigate` which I couldn't do before, is it for this reason that you passed navigation as a prop?

PR #204, opened 2024-05-18
fix: refine events

GustaveCharles:
# What I did

I did a few things related to the Event type and Event loading:
- Remove the Date type on the events. This is done in order for the event to be serializable and thus be able to be passed as props in some specific screens. Indeed, a React Date is not serializable.
- make the events load once added
- add a list of participants to the event type
- add the creator of the event in the event type

# How I did it
- I added a hook for refreshing the page on a swipe down from the user.
- When an event is loaded or fetched from the database, it is a string in the ISO format. When we want to display it, we create a Date object. I had to change the query selector too.

# How to verify it
For the swipe down, you can go to the Home page, create an event, and then swipe down as in the video. This should automatically display your event.

# Demo video

[

https://github.com/uniconnect-epfl/uniconnect/assets/91184289/caf00ccb-e763-4ff3-85ed-ac6ea69114c7

](url)
# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


Aidasvenc:
Could you explain the reason why you needed to change the Data type? You could also add it to the PR description for clarity

firebase/ManageEvents.ts
```diff
@@ -49,18 +47,16 @@ export const getAllPastEvents = async () => {
     const eventsRef = collection(db, "events")
 
     // Create a query against the collection.
-    // This query retrieves events where the eventDate is less than the current date and time.
-    const q = query(eventsRef, where("date", "<", new Date()), orderBy("date", "desc"))
+    const nowIsoString = new Date().toISOString()
```
Aidasvenc, line 50: Just a little detail - users in different time zones will have different current time returned by `new Date()`. Therefore, some events for such users may appear as still upcoming even though they are already done. This is just a problem that I have encountered in real life and unfortunately the only solution is to make these checks in the backend. This unfortunately cannot be done in Firebase.

screens/Maps/EventMap.tsx
```diff
@@ -26,6 +26,15 @@ const EventMap = () => {
...
 
+  const computeEventDate = (date: string) => {
+    const eventDate = new Date(date).toLocaleDateString("en-US", {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric'
+    })
+    return eventDate
+  }
+
```
Aidasvenc, line 37: I can see this function being repeatedly used in different files. I suggest defining this function externally and importing it in each file where you use it.

Aidasvenc (commented):
Nice visual changes! Just some minor details remain to be changed and it will be good

firebase/ManageEvents.ts
GustaveCharles: Interesting! I did not have that perspective, such a pain in the ass haha

GustaveCharles:
> Could you explain the reason why you needed to change the Data type? You could also add it to the PR description for clarity

Sure, I've just added some info

Fyuryy (approved):
LGTM

PR #224, opened 2024-05-20
feat: search bar for announcement and recommandation system

alexandre-mrt:
# What I did

Implemented the search bar for the announcement.
Implemented the recommandation system. Depending on the user interests it order the announcements. 
Moreover if all 3 tags of the announcement are matching your interest it is shown by an indicator.

# How I did it
I implemented the search bar, using states in react.
I used map and sort function on an array of announcement to order them depending on user preferences.

# How to verify it
Go to the announcement screen and use the search bar.
If you change your Interests you will see that the recommandations will be different

# Demo video
image

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


GaelCondeLosada (changes requested):
Hey good PR, everything works on my android device but there are a few things that could be improved. Also I couldn't test the recommendation system because I could not update my interests.

PR #226, opened 2024-05-20
fix: save pin position

GaelCondeLosada:
# What I did
Addressed the problem of the pin position not being saved for events and always being Besançon by default. Now the pin position choosed will be saved and appear on the event page.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/88dc6f42-31a1-4ded-85fb-f4257a20c12f

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/screens/EventCreation/EventCreationScreen.tsx
+++ b/screens/EventCreation/EventCreationScreen.tsx
@@ -72,6 +72,7 @@ const EventCreationScreen = ({ navigation, isAnnouncement }: EventCreationScreen
     if(!point){
       console.log("Need a point to create an event")
       setPoint(undefined)
+      console.log("Event not created")
       showErrorToast("You must enter a location for an event")
       return
     }


screens/EventCreation/EventCreationScreen.tsx
```diff
@@ -69,7 +69,14 @@ const EventCreationScreen = ({ navigation, isAnnouncement }: EventCreationScreen
       showErrorToast("You must be logged in to create an event")
       return
     }
-    const eventId = await createEvent(title, description, date.toISOString(),{ x: 47.238458, y: 5.984155 }, location, "imageUrl", userId)
+    const eventId = await createEvent(
+      title, description, 
+      date.toISOString(), 
+      point || { x: 47.238458, y: 5.984155 },
```
AlbertoCentonze, line 75: Do we fallback to that random place if the point can't be fetched?  i imagine this is due to typescript constraints but can't we should an `Alert` instead?

AlbertoCentonze (changes requested):
Nice job, just one little detail I'd like to understand.

screens/EventCreation/EventCreationScreen.tsx
GaelCondeLosada: Yepp you're right!

AlbertoCentonze (approved):

PR #229, opened 2024-05-21
feat: link map and location

GaelCondeLosada:
# What I did
Make it so the location name and the marker on the map are now linked. They now both come from the `SelectLocationScreen`. Here you can now choose a point and the location name will be auto-filled and you can also search a location by it's name and it will find it on the map. There is not address autocomplete in the input field, I'll look to implement it in another PR.

# How I did it
- Removed the text input for the location name from the location creation screen and put it on the point selection map screen.
- Implemented point to address and address point with two functions that automatically fill the other when one is filled.
- I couldn't use `react-native-google-places-autocomplete` because it required activating billing on our project which we didn't want to. I found a free alternative that is [nominatim](https://nominatim.openstreetmap.org/ui/search.html).
- Added dev dependency `jest-fetch-mock` to be able to mock the interactions with nominatim.
- Implemented tests for the location selection screen.

# How to verify it
Launch the app in expo go. Go create an event, try to add a location by selecting a point and by entering an address. Create the event and check that it then has the right location.

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/5c2a141f-81b2-4fd1-9994-7fe64d3fa742

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


Fyuryy:
Good PR. The code looks clean and the functionalities work on my phone. However, the maps are not charging so I can't actually test it. That's why I cannot approve.

Aidasvenc (approved):
Wow, nice one here!

PR #234, opened 2024-05-22
feat: added all the back arrows on screens that needed it

GaelCondeLosada:
# What I did
Added back arrows on all screen that were missing it.

# How I did it
- Added the component `BackArrow` in the screens.
- Adapted the tests.

# How to verify it
- Open the app and check that the back arrow is on all screens that have one on the figma and that the arrow works correctly.
- You can also run the tests.

# Demo video

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
+++ b/screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
@@ -14,6 +14,7 @@ import { getAuth } from "firebase/auth"
 import LoadingScreen from "../../Loading/LoadingScreen"
 import { getUserData, addFriend, removeFriend } from "../../../firebase/User"
 import { User } from "../../../types/User"
+import { BackArrow } from "../../../components/BackArrow/BackArrow"
 
 type RootStackParamList = {
   ExternalProfile: {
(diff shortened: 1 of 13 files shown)


screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
```diff
@@ -77,6 +78,7 @@ const ExternalProfileScreen = () => {
 
   return (
     <View style={styles.container}>
+      <BackArrow/>
```
alexandre-mrt, line 81: When testing with the app it doesn't render correctly. You should try putting it outside of the container. Moreover the position is absolute so take care about overlapping on layouts

alexandre-mrt (changes requested):
LGTM. Just for the external profile try to make the layout great again

screens/Profile/ExternalProfileScreen/ExternalProfileScreen.tsx
GaelCondeLosada: There is another PR opened to fix this problem (#243 )

alexandre-mrt (approved):
LGTM

PR #244, opened 2024-05-24
feat: improved layout and functionality of registration flow

GaelCondeLosada:
# What I did
- Cleaned the layout throughout the registration flow, made it consistent (background color, input field size).
- Made the layout not move when the keyboard goes up.
- Made some fields required (like first name for example), you can't continue without filling it. It will display a toast to warn you.
- Made it so the labels are persistent when moving back and coming back to the interest screen.

# How I did it
- Modified already existing screens and styles files.

# How to verify it
Launch the app in expo go, create a new account, check the layout and try to break the functionalities, skip steps and so on...

# Demo video

https://github.com/uniconnect-epfl/uniconnect/assets/93329823/1b3cef35-0d2a-4af6-9a1c-8c93f0978fde

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


screens/Registration/AuthenticationScreen/AuthenticationScreen.tsx
```diff
@@ -69,11 +70,17 @@ const AuthenticationScreen: React.FC = () => {
...
+    style={[
+      styles.container, 
+      {height: Dimensions.get('window').height + insets.bottom + insets.top}
+    ]}
+      onPress={() => Keyboard.dismiss()}>
       <View
         style={[
           styles.view,
           { paddingBottom: insets.bottom, paddingTop: insets.top },
+          {height: Dimensions.get('window').height + insets.bottom + insets.top}
```
Aidasvenc, line 83: Why did you specify height, was `flex: 1` or `height: "100%"` not enough for both styles?

screens/Registration/DescriptionScreen/DescriptionScreen.tsx
```diff
@@ -20,11 +21,17 @@ const DescriptionScreen: React.FC = () => {
...
+      style={[
+        styles.container, 
+        {height: Dimensions.get('window').height + insets.bottom + insets.top}
+      ]}
+      onPress={() => Keyboard.dismiss()}>
       <View
         style={[
           styles.container,
           { paddingBottom: insets.bottom, paddingTop: insets.top },
+          {height: Dimensions.get('window').height + insets.bottom + insets.top}
```
Aidasvenc, line 34: same question here

screens/Registration/InterestsScreen/InterestsScreen.tsx
```diff
@@ -130,11 +140,17 @@ const InterestsScreen = () => {
...
+      style = {[
+        styles.container,
+        {height: Dimensions.get('window').height + insets.bottom + insets.top}
+      ]}
+      onPress={() => Keyboard.dismiss()}>
       <View
         style={[
           styles.container,
           { paddingBottom: insets.bottom, paddingTop: insets.top },
+          {height: Dimensions.get('window').height + insets.bottom + insets.top}
```
Aidasvenc, line 153: same here

Aidasvenc (commented):
Looks good except from the `height` specification choice. Once resolved, will be perfect!

Gaellosada:
> Looks good except from the `height` specification choice. Once resolved, will be perfect!

It's to prevent the layout from moving when the keyboard goes up. I agree that it does not look super clean... I didn't find a better way to do it. Do you know how I can do that in a cleaner way ?

Aidasvenc:
> > Looks good except from the `height` specification choice. Once resolved, will be perfect!
> 
> It's to prevent the layout from moving when the keyboard goes up. I agree that it does not look super clean... I didn't find a better way to do it. Do you know how I can do that in a cleaner way ?

Since it is a very subtle problem and we don't have much time, even a non-clean solution is acceptable since it gets the job done! Thank you for the clarification.

Aidasvenc (approved):
Thank you for the clarifications. LGTM!

PR #247, opened 2024-05-24
feat: differentiate events by host and participants labels

Aidasvenc:
# What I did
Now each event card will show whether you are already registered to an event and whether you are a host of that event or not. Also, each card contains a small profile picture of each participant.

Bonus: fixed layout in events screen so that it can accommodate more event cards with less padding.

# How I did it
- Added needed styling and components
- Defined a function to fetch all user pictures from the backend
- Changed the layout

# How to verify it
Try to create an event or register to it.

# Demo
drawing

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested


GaelCondeLosada (approved):
LGTM, everything works on my android device (I checked the box).

PR #266, opened 2024-05-28
feat: minor adjustments

AlbertoCentonze:
# What I did
1. Added a default image when creating an event.
2. Restored interest selection functionalities that got lost because of a merge.
3. Amended registration flow.

# How I did it
1. For the default image I just used the URL of an image that matches the style of the app as the default value.
2. Copy pasted lost code for the interests selection.
3. `git bisect` to find the commit that broke it, then just change the id in the stack navigator to stop having conflicts. 

# How to verify it
1. log-out if you are logged-in.
2. create a new account.
3. creation should be successful.
4. try to create an event and select some interests.

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/navigation/Main/MainStackNavigator.tsx
+++ b/navigation/Main/MainStackNavigator.tsx
@@ -177,7 +177,7 @@ const MainStackNavigator: React.FC = () => {
               options={{ headerShown: false }}
             />
             <Stack.Screen
-              name="Interests"
+              name="EventInterests"
               component={InterestsScreen}
               options={{ headerShown: false }}
               initialParams={{ eventMode: true }}
--- a/screens/EventCreation/EventCreationScreen.tsx
+++ b/screens/EventCreation/EventCreationScreen.tsx
@@ -25,7 +25,8 @@ import { getAuth } from "firebase/auth"
 import { User } from "../../types/User"
 import { BackArrow } from "../../components/BackArrow/BackArrow"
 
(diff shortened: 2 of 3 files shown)


Fyuryy:
I tested the demo with your code and there are a few things to address.
1. Display interests when they get picked for an event under the Interests selected text
2. Don't force people to choose 3 interests.. Cap it at 3 and enforce 1 interest as a minimum. People should be able to choose 1, 2 or 3 interests. 
3. This one is a bit critical, so you might want to look at it first, but when you create the event and press Validate, you stay in the screen. If you click 10 times in validate, well it creates the same event 10 times.
Otherwise it should be good

AlbertoCentonze:
> * Display interests when they get picked for an event under the Interests selected text
> * Don't force people to choose 3 interests.. Cap it at 3 and enforce 1 interest as a minimum. People should be able to choose 1, 2 or 3 interests.

Will tackle this in a future PR since this PR contain a critical fix that should be shipped ASAP.



> 3\. This one is a bit critical, so you might want to look at it first, but when you create the event and press Validate, you stay in the screen. If you click 10 times in validate, well it creates the same event 10 times.
>     Otherwise it should be good

Done!

Fyuryy (approved):

Fyuryy:
Now that the critical fix is SHIPPED, you might wanna start fixing the smaller ones 👯

PR #268, opened 2024-05-29
fix: solve layout and announcement problems

alexandre-mrt:
# What I did
The app notifies the user when he has no events yet. 
I added the sticky headers for the events screen so we know if we are seeing upcoming events or past ones.
We can now see the selected interests when we create an announcement

# How I did it
I added an empty array for events in the creation method, so the query in the database is not throwing an error but retrieve an empty array. It's detected and a different message than "Error with your connection is shown". 

# How to verify it
You can check the event screen to see the sticky header. 
You can try to go to the event screen without any events to see the correct notification "No events yet"
When selecting the interests for an announcement, you can now see what you selected.
# Demo video
image
image
image

# Pre-merge checklist
- [x] work correctly
- [x] are fully tested

### DIFF
--- a/firebase/Registration.ts
+++ b/firebase/Registration.ts
@@ -47,7 +47,8 @@ export async function storeInitialUserData(uid: string, email: string, firstName
       location: location,
       description: description,
       selectedInterests: selectedInterests,
-      profilePicture: ""
+      profilePicture: "",
+      events : []
     })
   } catch (error) {
     showErrorToast("There was an error storing your user data, please try again.")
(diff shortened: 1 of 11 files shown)


firebase/Registration.ts
```diff
@@ -47,7 +47,8 @@ export async function storeInitialUserData(uid: string, email: string, firstName
       location: location,
       description: description,
       selectedInterests: selectedInterests,
-      profilePicture: ""
+      profilePicture: "",
+      events : []
```
GustaveCharles, line 51: I think that we don't want this, because the context doesn't allow for it. Are you sure we need it?

alexandre-mrt: We need it to avoid an error, fetching null events. Now we can differentiate if it's null or just empty

GaelCondeLosada (approved):
LGTM don't forget to address sonar cloud issues.

